PR #251, opened 2024-03-28
Docker file with smile serde

azatsafin:

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

### DIFF
--- a/api/Dockerfile-Serdes
+++ b/api/Dockerfile-Serdes
@@ -1,0 +1,15 @@
+ARG BASE_UI_IMAGE=ghcr.io/kafbat/kafka-ui:main
+FROM $BASE_UI_IMAGE
+
+ARG SERDE_SMILE_JAR=https://github.com/kafbat/ui-serde-smile/releases/download/v1.0.1/serde-smile-v1.0.1-jar-with-dependencies.jar
+ARG SERDE_SMILE_VERSION=1.0.1
+USER root
+RUN mkdir /usr/lib/serdes; \
+    chown -R kafkaui:kafkaui /usr/lib/serdes
+USER kafkaui
+
+RUN cd /usr/lib/serdes; \
+    wget -O "kafka-ui-smile-serde-${SERDE_SMILE_VERSION}.jar" $SERDE_SMILE_JAR
+ENV kafka.clusters.0.serde.0.name="Smile"
+ENV kafka.clusters.0.serde.0.filePath="/usr/lib/serdes/kafka-ui-smile-serde-${SERDE_SMILE_VERSION}.jar"
+ENV kafka.clusters.0.serde.0.className="io.kafbat.serde.smile.SmileSerde"


api/Dockerfile-Serdes
```diff
@@ -0,0 +1,15 @@
+ARG BASE_UI_IMAGE=ghcr.io/kafbat/kafka-ui:main
+FROM $BASE_UI_IMAGE
+
+ARG SERDE_SMILE_JAR=https://github.com/kafbat/ui-serde-smile/releases/download/v1.0.1/serde-smile-v1.0.1-jar-with-dependencies.jar
```
germanosin, line 4: Could we add glue serde also?

germanosin (changes requested):

api/Dockerfile-Serdes
Haarolean: @germanosin yes, it's a WIP

api/Dockerfile-Serdes
```diff
@@ -0,0 +1,38 @@
+ARG BASE_UI_IMAGE=ghcr.io/kafbat/kafka-ui:main
```
Haarolean, line 1: This means that if we run this workflow without running our main image, we're gonna build an outdated image. We need to either depend on the base image workflow or use some other approach.

api/build-with-serders.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+# Smile
+smile_latest_release=$(curl -s "https://api.github.com/repos/kafbat/ui-serde-smile/releases/latest")
```
Haarolean, line 3: Can we collapse this script even more so we can have a list of repos (like kafbat/ui-serde-smile) so we could iterate over it and pull each jar?

api/Dockerfile-Serdes
```diff
@@ -0,0 +1,38 @@
...
+
+# Install Smile Serde
+ARG SERDE_SMILE_JAR=https://github.com/kafbat/ui-serde-smile/releases/download/v1.0.1/serde-smile-v1.0.1-jar-with-dependencies.jar
+ARG SERDE_SMILE_VERSION=1.0.1
+
+RUN cd /usr/lib/serdes; \
+    wget -O "kafka-ui-smile-serde-${SERDE_SMILE_VERSION}.jar" $SERDE_SMILE_JAR
+
+ENV kafka.clusters.0.serde.0.name="Smile"
+ENV kafka.clusters.0.serde.0.filePath="/usr/lib/serdes/kafka-ui-serde-smile-${SERDE_SMILE_VERSION}.jar"
```
Haarolean, line 17: @kafbat/backend need to discuss this, perhaps we could introduce a way to load serdes from one common path so we don't have to add these many properties and delegate the rest (like serde properties) to the user

.github/workflows/build-serders.yml
```diff
@@ -0,0 +1,70 @@
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get latest Serders
+        id: get_serders
+        run: |
+          ./api/build-with-serders.sh
```
Haarolean, line 50: btw it's "serdes", not "serders" :)

Haarolean (changes requested):

api/Dockerfile-Serdes
azatsafin: it is possible, but I think it is better approach
this enables to track versioning of serders separatly from the kafka-ui main.

PR #261, opened 2024-04-02
BE: Fix KafkaConsumerGroupTests on Windows

busches:

`shouldOkWhenConsumerGroupIsNotActive` was returning a 400 instead of a 200 without the try with resources block, once I added it, the test consistently passes. Additionally removed some unused variables and used `toList()` added with JDK 17.

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks


germanosin:
Hi, @busches. Thanks for your contribution! Looks nice! @iliax don't you mind to review it?

iliax:
@busches your changes looks fine, but I can't understand how they can fix tests on windows) Can you please add more context here and describe why this is working now?

busches:
What the code change does is ensure that the Consumer Group is cleaned up, with the try with resources, before it executes the API request. The same is true for the other test, where we leave the Consumer Group up until after the API request. That test did not have any issues passing for me, but I changed it to mirror the other for self documenting purposes.

The issue may not be related to Windows at all, but that's what I'm running. It could be a case of my computer is too slow to remove the consumer group fully before the API request runs or potentially something with my JDK, I was using Oracle and will swap to Termium tonight. Note: there were three failing test suites I encountered, this one, one for the Protobuf stuff (unsure of the cause) and another that is 100% OS line endings that I'll fix next.

Haarolean (approved):

busches:
@Haarolean I would like to fix the other test failures I'm seeing on Windows, but no ETA on when I can get them in :)

busches:
Everything is passing now on Windows. I didn't like that I had to change `ProtobufFileSerde`, the other option is to change the test  files to not be nested in the `language` folder.

api/src/main/java/io/kafbat/ui/serdes/builtin/ProtobufFileSerde.java
```diff
@@ -408,7 +408,7 @@ private Map<String, ProtoFile> loadFilesWithLocations() {
         files.filter(p -> !Files.isDirectory(p) && p.toString().endsWith(".proto"))
             .forEach(path -> {
               // relative path will be used as "import" statement
-              String relativePath = baseLocation.relativize(path).toString();
+              String relativePath = baseLocation.relativize(path).toString().replace("\\", "/");
```
Haarolean, line 411: Could we get rid of this and possibly replace it with some native java nio API calls if required?

Haarolean (changes requested):

api/src/main/java/io/kafbat/ui/serdes/builtin/ProtobufFileSerde.java
busches: What are you looking to get rid of? The goal of this PR wasn't to rewrite this function.

Haarolean: Yes, but "fixing tests on windows" doesn't imply altering the production code. I believe this could lead to unexpected side effects in production, perhaps on non-windows OS.

busches: That is fair. I can't say I understand this implementation well enough to know why one part uses the OS separator and another doesn't, which causes the mismatch. This part is using the OS separator, and I would wager rewriting it will not solve the problem. 

I can revert this portion if you like, but the E2E tests are passing on multiple OSes now and the slash being replaced is only used by Windows.

Haarolean: I would prefer not to have any hacky solutions in production code to be able to run tests on some weird environments.
I believe this can be somehow solvable with standard APIs, but I don't have time to dig into this.
I'd be happy to accept a PR which can alter this production code but in a way which doesn't introduce OS-specific hacks. Otherwise, at this point, let's run the tests in wsl or docker until we don't have a better solution.

Haarolean (approved):

Haarolean:
Thanks for a contribution and sorry for the delay

Issue #376, opened 2024-05-09
Infra: Ship ARM images


hugomiguelabreu:
### Issue submitter TODO list

- [X] I've looked up my issue in [FAQ](https://ui.docs.kafbat.io/faq/common-problems)
- [X] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [X] I've tried running `main`-labeled docker image and the issue still persists there
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md)

### Describe the bug (actual behavior)

Whilst running the current `ghcr.io/kafbat/kafka-ui` image in a ARM based laptop with [podman](https://github.com/containers/podman), there is a unexpected error:
```
[Too many errors, abort]
```

This is due to having a ARM images built based of a sha256 explicit AMD image. Therefore, for the kafka-ui image, a base AMD image is being (zulu) used which is leading to this issue. 

### Expected behavior

Kafka UI container image runs out-of-the-box in both ARM and AMD architectures.

### Your installation details

```
docker run -it -p 8080:8080 ghcr.io/kafbat/kafka-ui
```

### Steps to reproduce

Have a ARM based laptop (e.g. M1) and install [podman](https://github.com/containers/podman) with docker integration.
After the setup, it is not possible to run the Kafka UI image due to AMD binaries in the base image.

### Logs

```
Standard Commons Logging discovery in action with spring-jcl: please remove commons-logging.jar from classpath in order to avoid potential conflicts































[Too many errors, abort]
```

Haarolean:
Hey, we're indeed planning to ship ARM images, but as a sidenote, our image works fine for me on apple silicon, but I use docker (orbstack to be precise) rather then podman.

hugomiguelabreu:
Indeed! I also tested with Colima - https://github.com/abiosoft/colima - and it works as expected, though not with podman, after some investigation I assumed it is related with Rosetta support (podman rosetta support is still ongoing: https://github.com/containers/podman/pull/21670)

Nevertheless I have opened the PR - https://github.com/kafbat/kafka-ui/pull/378, this at least should ensure that the current shipped ARM image is based from a ARM Azul Java image

PR #378, opened 2024-05-10
Infra: Ensure image arch automatic inference

hugomiguelabreu:

Changes the base image used to ensure automatic inference, guaranteeing a correct base image for the built kafka-ui one.
Building the image with the updated base one the container boots successfully.

image

Closes: https://github.com/kafbat/kafka-ui/issues/376 

- Build and Run locally.

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ To run Kafbat UI, you can use either a pre-built Docker image or build it (or a 
 ## Quick start (Demo run)
 
 ```
-docker run -it -p 8080:8080 -e DYNAMIC_CONFIG_ENABLED=true image: ghcr.io/kafbat/kafka-ui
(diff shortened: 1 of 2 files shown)


hugomiguelabreu:
I am not sure about the image used as I wasn't able to infer the actual version from the hash. I based off the latest one, locking the hotfix.

Let me know if we need some other particular version to be compliant with any active CVE

paxadax (commented):
Thank you , this will be really helpful

Haarolean:
> I am not sure about the image used as I wasn't able to infer the actual version from the hash. I based off the latest one, locking the hotfix.
> 
> Let me know if we need some other particular version to be compliant with any active CVE

the point was pinning the version, but your solution is better, I guess, since the minor version won't get updated either :) This should work generally, unless we have to bump the version before a minor is released to circumvent any CVEs in alpine. Thanks!

Haarolean (approved):

PR #380, opened 2024-05-11
FE: Fix refetching data on window focus

### LINKED ISSUE #348: FE: Do not refetch data on window focus
Some components like `ui/clusters/local/schemas/test/edit` does refetch data via tsquery on window focus.
This is considered a harmful behavior, because if there are any unsaved changes made to the schema itself the data will be refetched and the changes would be lost.

TODO:
1. Check all the places where window focus refetch is on but shouldn't be
2. Add `QUERY_REFETCH_OFF_OPTIONS` in these places

masonwoodford:

Closes #348
I created a new constant QUERY_REFETCH_ON_WINDOW_FOCUS_OFF_OPTION that just has refetchOnWindowFocus false rather than

export const QUERY_REFETCH_OFF_OPTIONS = {
refetchOnMount: false,
refetchOnWindowFocus: false,
refetchIntervalInBackground: false,
};

I created this new constant because when all 3 options are provided, even when the Edit field is updated the UI will not reflect the changes until the page is refreshed. I added this constant to the two Edit components that fetch and place into editable fields.

- [x] No need to
- [x] Unit checks
- [x] Integration checks

image

### DIFF
--- a/frontend/src/lib/constants.ts
+++ b/frontend/src/lib/constants.ts
@@ -73,8 +73,9 @@ export const QUERY_REFETCH_OFF_OPTIONS = {
   refetchIntervalInBackground: false,
 };
 
-export const QUERY_REFETCH_ON_WINDOW_FOCUS_OFF_OPTION = {
+export const QUERY_REFETCH_LIMITED_OPTIONS = {
   refetchOnWindowFocus: false,
+  refetchIntervalInBackground: false,
 };
 
 // Cluster Form Constants
--- a/frontend/src/lib/constants.ts
+++ b/frontend/src/lib/constants.ts
@@ -73,6 +73,10 @@ export const QUERY_REFETCH_OFF_OPTIONS = {
   refetchIntervalInBackground: false,
 };
 
+export const QUERY_REFETCH_ON_WINDOW_FOCUS_OFF_OPTION = {
+  refetchOnWindowFocus: false,
+};
+
 // Cluster Form Constants
 export const AUTH_OPTIONS = [
   { value: 'SASL/JAAS', label: 'SASL/JAAS' },
--- a/frontend/src/lib/hooks/api/topics.ts
+++ b/frontend/src/lib/hooks/api/topics.ts
@@ -4,7 +4,12 @@ import {
   consumerGroupsApiClient,
   messagesApiClient,
 } from 'lib/api';
-import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
+import {
+  useMutation,
+  useQuery,
+  useQueryClient,
+  UseQueryOptions,
+} from '@tanstack/react-query';
 import {
   CreateTopicMessage,
   GetTopicDetailsRequest,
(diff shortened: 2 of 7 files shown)


frontend/src/lib/constants.ts
```diff
@@ -73,6 +73,10 @@ export const QUERY_REFETCH_OFF_OPTIONS = {
   refetchIntervalInBackground: false,
 };
 
+export const QUERY_REFETCH_ON_WINDOW_FOCUS_OFF_OPTION = {
+  refetchOnWindowFocus: false,
```
Haarolean, line 77: I guess we'd need `refetchIntervalInBackground: false` here as well

frontend/src/lib/hooks/api/topics.ts
```diff
@@ -50,10 +51,18 @@ export function useTopics(props: GetTopicsRequest) {
   );
 }
 export function useTopicDetails(props: GetTopicDetailsRequest) {
-  return useQuery(topicKeys.details(props), () => api.getTopicDetails(props));
+  return useQuery(
+    topicKeys.details(props),
+    () => api.getTopicDetails(props),
+    QUERY_REFETCH_ON_WINDOW_FOCUS_OFF_OPTION
```
Haarolean, line 57: the hooks themselves might be used in multiple different places. I don't see a problem with refetching the defails when e.g. viewing topic details (which is okay since you can't edit anything in there), but this should have a differen behavior in topic *edit* view, can we adjust it like that?

Haarolean (changes requested):

masonwoodford:
Have updated with the requested changes by passing a parameter to the query functions. I did want to point out that even though the Edit topic component fetches on window focus I don't see it having the issue that Schemas has, but for consistency I added it there but can remove it as well. It also removes unnecessary api requests.

masonwoodford:
Sorry about that didn't mean to re request review didn't see you had requested another review.

frontend/src/lib/constants.ts
```diff
@@ -73,6 +73,11 @@ export const QUERY_REFETCH_OFF_OPTIONS = {
   refetchIntervalInBackground: false,
 };
 
+export const QUERY_REFETCH_LIMITED_OPTIONS = {
```
Mgrdich, line 76: if we are going to apply the refetch of all queries on window focus , it is better to provide the parameter and the config to the root `QueryClient` that gets pass to the provider rather than passing it as an option to all the queries.

FYI @Haarolean

masonwoodford: Currently it is not being passed to all queries just ones that are editable so that the fetching does not overwrite any temporary changes made. Should this be provided to all queries instead?

Haarolean: Nope, as we discussed, only for the views which support editable parts: [source](https://github.com/kafbat/kafka-ui/issues/348#issuecomment-2094976884)

cc @Mgrdich

Mgrdich (approved):

Haarolean:
@masonwoodford thank you and sorry for the delay!

PR #443, opened 2024-06-13
fix Infra: Do not run e2e reports task on forks #313

### LINKED ISSUE #313: Infra: Do not run e2e reports task on forks
There's a repo name check but it doesn't seem to work, need to check for the fork signs in another way
https://github.com/kafbat/kafka-ui/actions/runs/8763118233/job/24051917042?pr=308

agungardiyanta:

Resolves #313

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks


.github/workflows/e2e-run.yml
```diff
@@ -134,7 +134,7 @@ jobs:
   reports:
     runs-on: ubuntu-latest
     needs: tests
-    if: ${{ !cancelled() && github.repository == 'kafbat/kafka-ui' }}
+    if: ${{ !cancelled() && github.repository == 'kafbat/kafka-ui' && github.event.pull_request.head.repo.full_name == github.repository }}
```
Haarolean, line 137: do we even need `github.repository == 'kafbat/kafka-ui'` anymore?

Haarolean (changes requested):

.github/workflows/e2e-run.yml
agungardiyanta: oh ya we don't need it anymore, I will fix this and push the changes

.github/workflows/e2e-run.yml
```diff
@@ -134,7 +134,7 @@ jobs:
   reports:
     runs-on: ubuntu-latest
     needs: tests
-    if: ${{ !cancelled() && github.repository == 'kafbat/kafka-ui' }}
+    if: ${{ !cancelled() && github.event.pull_request.head.repo.full_name == github.repository }}
```
Haarolean, line 137: One more thing, out of three workflows we have which call this one, only one of them has `pull_request` trigger. How will this work for two others (`workflow_dispatch` and `schedule`)?

agungardiyanta: it will not run the reports task, do we want to make the task still run for two other workflow?

agungardiyanta: I already change the condition, so now the reports task will run if not cancelled and get trigger not from pull request org trigger from pull request that not from forks

.github/workflows/e2e-run.yml
```diff
@@ -134,7 +134,7 @@ jobs:
   reports:
     runs-on: ubuntu-latest
     needs: tests
-    if: ${{ !cancelled() && github.repository == 'kafbat/kafka-ui' }}
+    if: ${{ !cancelled() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
```
Haarolean, line 137: This won't work as this will always be `workflow_dispatch`, as it's the only trigger for this particular workflow. 
I'm not sure what's the best way to do this properly, but the way we currently do this is that we forward the original `event_name` via workflow `inputs` (see `backend_tests.yml` for an example)

Haarolean (changes requested):

.github/workflows/e2e-run.yml
agungardiyanta: so we only want this Reports task only run when the workflow triggers by workflow_dispatch (manually)?

Haarolean: No, we still have PRs which are not from forks.

`github.event_name` here always would be `workflow_dispatch`, so as we'd need an original trigger name, we need to either forward it the way I described or some other way.

agungardiyanta: Oh if that is the concern my current changes already handle it because I put `OR`/ `||`operation so it will still run for internal `pull request`

`${{!cancelled() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}`

you can check my testing repository 
Testing [Internal PR](https://github.com/agungardiyanta/docker-sample/actions/runs/9562700738/job/26359619835) 
image

Testing [workflow_dispatch](https://github.com/agungardiyanta/docker-sample/actions/runs/9562676574/job/26359552343)
image

Testing Forks [Outside PR](https://github.com/agungardiyanta/docker-sample/actions/runs/9513452373/job/26344324400) 
image

Haarolean: Once again, the `github.event_name` var will _never_ be anything but `workflow_dispatch` in this case, no matter which workflow calls the current one, this renders your if condition invalid. You can verify this by calling the workflow from three other caller workflows and dumping the github context into logs. 

We need to forward the original event name from the caller workflow (or use any other mechanism if you're aware of ones).

agungardiyanta: Hi, @Haarolean I did a little experiment in my repo to test what the `github.event_name` will be received by template workflow `e2e.yaml`, and it seems my workflow can get the original event_name without adding a new parameter CMIIW. 
[workflow_dispatch](https://github.com/agungardiyanta/docker-sample/blob/master/.github/workflows/manual.yml)
image

[pull request](https://github.com/agungardiyanta/docker-sample/blob/master/.github/workflows/pr.yaml)
image

Haarolean: Hi, sorry for the delay. That's surprising, as the input with the original event name in backend tests workflow was added because I couldn't get the event name the same way you did in your example.

Haarolean:
I also had a concern related to `event_name != pull_request` but with a closer look I see that we don't call e2e workflow from pull_request_target events anymore.

Haarolean (approved):

agungardiyanta:
Thank you! I'm happy to contribute to Kafbat UI.

Issue #440, opened 2024-06-12
FE: Topics: Remember user polling options


Haarolean:
Polling options changed by the user should be persisted in localstorage per topic.

Settings to consider:
1. Polling mode (oldest/newest/etc)
2. Key/value serdes
3. Active filter(-s)
4. Search box text field
5.

Mgrdich:
Local storage and the URL can conflict with each other we must determine which one has a priority and is the source of truth

Nilumilak:
> Local storage and the URL can conflict with each other we must determine which one has a priority and is the source of truth @Haarolean

I think, that in my case the URL will have the priority. That is what I discussed with @Haarolean

PR #453, opened 2024-06-23
FE: Topics: Remember user polling options

Nilumilak:
Resolves: #440


Leshe4ka (approved):

Leshe4ka (approved):

frontend/src/components/common/Search/Search.tsx
```diff
@@ -31,6 +31,12 @@ const Search: React.FC<SearchProps> = ({
   const [searchParams, setSearchParams] = useSearchParams();
   const ref = useRef<ComponentRef<'input'>>(null);
 
+  useEffect(() => {
+    if (ref.current != null && value) {
```
Mgrdich, line 35:
```suggestion
    if (ref.current !== null && value) {
```

frontend/src/components/common/Select/Select.tsx
```diff
@@ -42,6 +42,10 @@ const Select = <T extends object>(
   const [selectedOption, setSelectedOption] = useState(value);
   const [showOptions, setShowOptions] = useState(false);
 
+  useEffect(() => {
```
Mgrdich, line 45: what is the case where the value is changed from the outside that we need this effect

frontend/src/lib/hooks/useMessagesFiltersFields.ts
```diff
@@ -0,0 +1,94 @@
...
+  }>('message-filters-fields', {});
+
+  const removeMessagesFilterFields = () => {
+    setMessageFilters((prev) => {
+      const { [topicName]: topicFilters, ...rest } = prev || {};
+      return rest;
+    });
+  };
+
+  const removeMessagesFiltersField = (key: keyof MessagesFilterFieldsType) => {
```
Mgrdich, line 30:
```suggestion
  const removeMessagesFiltersField = (key: MessagesFilterFieldsType) => {
```

because keys and values are the same we can use it like this

Nilumilak: but then typescript won't like it
image

frontend/src/components/common/Select/Select.tsx
Nilumilak: without this useEffect, filters won't be applied when url params are empty and we take them from localStorage

frontend/src/components/common/Search/Search.tsx
Nilumilak: fixed

Mgrdich (approved):

PR #509, opened 2024-07-29
Serde: Enable header modifications by custom serializer

### LINKED ISSUE #346: Serde: Enable header modifications by custom serializer
### Issue submitter TODO list

Some schema registries allow placing schema details in message headers instead of the payload. For example, Apicurio Registry serializer enables writing the artifact identifier to message headers. 

However, when creating custom serializers that implement the `io.kafbat.ui.serde.api.Serde` interface with `io.kafbat.ui.serde.api.Serde.Serializer`, these interfaces don't provide direct access to message headers, which can be limiting.

Expend `io.kafbat.ui.serde.api.Serde.Serializer` to support message headers editing.

The least complicated solution would be to extend the serializer interface with additional method:
```java
default byte[] serialize(String input, Headers headers) {
    return serialize(input);
}
```

Where `Headers` is `org.apache.kafka.common.header.Headers`, a mutable collection of headers.

If immutable headers are required (e.g., `io.kafbat.ui.serde.api.RecordHeaders`), then the interface might be extended with this method:
```java
default SerializeResult serialize(String input, RecordHeaders headers) {
    return new SerializeResult(serialize(input), headers);
}
```

Where `SerializeResult` would be a new class that contains serialized message and headers.

By adding this method, `io.kafbat.ui.serde.api.Serde.Serializer` would move one step closer to Kafka's `org.apache.kafka.common.serialization.Serializer` interface.

There is no alternative for this feature. For instance, when using Apicurio Registry, if a serializer includes schema details directly in the message payload, it prepends a magic byte to the beginning of the message. If the deserializer doesn't expect this magic byte, it will not be able to read the message. So, to use Kafka-UI to produce messages with Apicurio Registry, it is necessary to modify message headers in the serializer.

### Version you're running

2956664

### Additional context

_No response_

MDzaja:

Closes #346.
Extends `io.kafbat.ui.serde.api.Serde.Serializer` to support message headers editing by custom serializers.

- [x] No need to
A custom serializer was developed, and the feature was tested with it.
- [x] Unit checks
- [x] Integration checks

maki (4)


api/src/main/java/io/kafbat/ui/serdes/SerdeInstance.java
```diff
@@ -78,10 +80,20 @@ public boolean canDeserialize(String topic, Serde.Target type) {
   }
 
   public Serde.Serializer serializer(String topic, Serde.Target type) {
-    return wrapWithClassloader(() -> {
-      var serializer = serde.serializer(topic, type);
-      return input -> wrapWithClassloader(() -> serializer.serialize(input));
-    });
+    var serializer = serde.serializer(topic, type);
+    // Create a dynamic proxy instance for the Serde.Serializer interface
+    return (Serde.Serializer) Proxy.newProxyInstance(
```
Haarolean, line 85: why we'd need a proxy here?

Haarolean (changes requested):

api/src/main/java/io/kafbat/ui/serdes/SerdeInstance.java
MDzaja: We need a proxy here because the `Serializer` interface now has multiple methods, so we cannot use a simple lambda for wrapping (as was done previously). A proxy provides a **generic**, **scalable**, and **clean solution** to wrap all methods of the `Serializer` interface dynamically. This ensures that every method invocation on the interface is consistently wrapped with the necessary logic (e.g., `wrapWithClassloader`).

While it’s possible to use an explicit or anonymous class implementation to achieve the same goal, using a proxy reduces boilerplate and makes the code easier to maintain in the long run. For comparison, here’s how an anonymous wrapper class could look:

``` java
public Serde.Serializer serializer(String topic, Serde.Target type) {
    return wrapWithClassloader(() -> {
        var serializer = serde.serializer(topic, type);
        return new Serde.Serializer() {
            @Override
            public byte[] serialize(String input) {
                return wrapWithClassloader(() -> serializer.serialize(input));
            }

            @Override
            public byte[] serialize(String input, Headers headers) {
                return wrapWithClassloader(() -> serializer.serialize(input, headers));
            }
        };
    });
}
```

While this approach works, it introduces repetitive boilerplate for every method, which becomes harder to manage as the interface evolves.

marcosschroh:
Is there any news of this? We are facing a similar problem. We want serialize the value according to some `headers` values.

Haarolean:
> Is there any news of this @MDzaja ? We are facing a similar problem. We want serialize the value according to some `headers` values.

we'll take a look, thanks for bumping

marcosschroh:
Hi, Is there anything that we can do to help you with this feature?

api/src/main/java/io/kafbat/ui/serdes/SerdeInstance.java
Haarolean: We've replaced the proxy with an anonymous class call no avoid stacking spring aop + our proxy on each call

Haarolean:
@MDzaja Thanks for your first contribution to kafbat ui! Sorry it took this long

marcosschroh:
Are you going to release this feature in https://central.sonatype.com/artifact/io.kafbat.ui/serde-api?

Haarolean:
@marcosschroh #1243

Issue #324, opened 2024-04-28
Title of page matches the contents of the page.


camjmayhew:
### Issue submitter TODO list

- [X] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md) and the feature is not present there

### Is your proposal related to a problem?

Whenever I have multiple tabs of Kafka UI open, I find it hard to navigate between them / choose the correct one when searching for a specific tab, as they are all called the same thing: "UI For Apache Kafka". It would be helpful to have some extra information in the title.

### Describe the feature you're interested in

Each tab has a title relating to its contents. 
E.g if the url is .../kafkaui-public/ui/clusters/staging/all-topics?perPage=25, the title could be All Topics - Staging. 
Other clear examples would be Schema Registry, Consumers, Brokers etc. 
Additionally, having better titles for Topics would be super helpful when on a specific topic and tab. 
E.g (TopicName) - Messages/Consumers/Settings. 

### Version you're running

0.6

### Additional context

Posted this question on the discord and was directed to submit it as a ticket here.

PR #540, opened 2024-09-11
Fixed active sidebar menu is highlighted for the correct Kafka instance

AlexisD02:

This pull request addresses a bug in Kafka UI where it was not clear which Kafka instance was being displayed. Previously, the sidebar options (e.g., Topics, Brokers, Consumers) were highlighted for all Kafka instances, causing confusion. 

**Changes Include:**
- Updated the page header to display '<instance name> Topics' instead of just "Topics" to clearly indicate which Kafka instance is being viewed.
- Ensured that the selected menu option is highlighted only for the Kafka instance currently being displayed, improving the clarity of navigation.

Please review the changes to ensure:
- The page header correctly reflects the current Kafka instance.
- The sidebar highlights only the selected menu option for the active Kafka instance.

### DIFF
--- a/frontend/src/components/ConsumerGroups/Details/ResetOffsets/ResetOffsets.tsx
+++ b/frontend/src/components/ConsumerGroups/Details/ResetOffsets/ResetOffsets.tsx
@@ -38,6 +38,7 @@ const ResetOffsets: React.FC = () => {
   return (
     <>
       <PageHeading
+        clusterName={routerParams.clusterName}
         text={consumerGroupID}
         backTo={clusterConsumerGroupsPath(routerParams.clusterName)}
         backText="Consumers"
--- a/frontend/src/components/Nav/ClusterMenu/ClusterMenu.tsx
+++ b/frontend/src/components/Nav/ClusterMenu/ClusterMenu.tsx
@@ -31,7 +31,7 @@ const ClusterMenu: FC<ClusterMenuProps> = ({
 }) => {
   const location = useLocation();
 
-  const getIsMenuItemActive = (path: string) => location.pathname === path;
+  const getIsMenuItemActive = (path: string) => location.pathname.includes(path);
 
   const hasFeatureConfigured = (key: ClusterFeaturesEnum) =>
     features?.includes(key);
--- a/frontend/src/components/Nav/ClusterMenu/ClusterMenu.tsx
+++ b/frontend/src/components/Nav/ClusterMenu/ClusterMenu.tsx
@@ -1,23 +1,16 @@
-import React, { type FC, useState } from 'react';
+import React, { FC } from 'react';
 import { Cluster, ClusterFeaturesEnum } from 'generated-sources';
 import * as S from 'components/Nav/Nav.styled';
 import MenuTab from 'components/Nav/Menu/MenuTab';
 import MenuItem from 'components/Nav/Menu/MenuItem';
 import {
   clusterACLPath,
-  clusterAclRelativePath,
-  clusterBrokerRelativePath,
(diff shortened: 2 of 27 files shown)


Haarolean (changes requested):
I don't get how is this related to #324 which is about tab's title?

AlexisD02:
> I don't get how is this related to #324 which is about tab's title?

Thanks for pointing this out! I looked for the issue related to my task but couldn't find anything. Could you point me to the correct issue if I'm missed something?

Haarolean:
> > I don't get how is this related to #324 which is about tab's title?
> 
> Thanks for pointing this out! I looked for the issue related to my task but couldn't find anything. Could you point me to the correct issue if I'm missed something?

I believe there's no issue related to these changes. I haven't experienced the issue you're describing either, as this is supposed to be a bugfix, let's file a bug report first, as I'm not being able to reproduce this.
image

hotsezus:
I'm experiencing the same issue using `main` tagged image from docker hub
image

Haarolean:
@hotsezus any specific steps to reproduce?

hotsezus:
> @hotsezus any specific steps to reproduce?

Nothing special. Just use `kafbat/kafka-ui:e7df880cdebe3977d06edee54a538bcaf9416649` image setting up connection to multiple clusters using `config.yaml`:
```yml
services:
  kafbat-ui:
    image: kafbat/kafka-ui:e7df880cdebe3977d06edee54a538bcaf9416649
    restart: on-failure
    ports:
      - 8080:8080
    environment:
      DYNAMIC_CONFIG_ENABLED: 'true'
    volumes:
      - ./config.yml:/etc/kafkaui/dynamic_config.yaml
```

frontend/src/components/ConsumerGroups/Details/ResetOffsets/ResetOffsets.tsx
```diff
@@ -38,6 +38,7 @@ const ResetOffsets: React.FC = () => {
   return (
     <>
       <PageHeading
```
Mgrdich, line 40: Instead of giving the Pageheading cluaterName couldn't we just use the hook param inside and get it there since we are giving it everywhere

Haarolean: @AlexisD02 could you address this?

Haarolean: @Mgrdich PTAL

frontend/src/components/Nav/ClusterMenu/ClusterMenu.tsx
```diff
@@ -27,18 +20,17 @@ interface ClusterMenuProps {
   name: Cluster['name'];
   status: Cluster['status'];
   features: Cluster['features'];
-  singleMode?: boolean;
+  openTab?: string | false;
```
Mgrdich, line 23:
```suggestion
  openTab?: string;
```

do we need the `false` can't we treat `undefined` as false

Mgrdich: and maybe make the tabs as a `enums`  in terms of object with `as const` so that we don't pass any arbitrary string

Haarolean: @Mgrdich these are cluster names, not the menu items, so an enum is not applicable
`handleTabChange(cluster.name)}`

Haarolean:
Should be fixed within #1120 and/or #984

PR #541, opened 2024-09-11
FE: CG: Fix names display overflow issue

### LINKED ISSUE #528: UX: Adjustable column sizes
### Issue submitter TODO list

_No response_

I would like to be able to make the dashboard column larger or smaller.
The problem is that the names of the clusters are too long and are not displayed in full, making it difficult to select the right cluster 

_No response_

### Version you're running

kafbat UI 2956664 v1.0.0

### Additional context

_No response_

AlexisD02:

This pull request addresses a bug in Kafka UI where consumer group IDs sometimes appear truncated due to insufficient space for display. The issue affects the readability of consumer group names on the consumer groups page.

**Changes Include:**
- Improved the layout of consumer group IDs to ensure they are displayed correctly based on the available space.

Please review the changes to confirm:
- Consumer group names are fully visible and properly formatted according to the available space.

Fixes #528

### DIFF
--- a/frontend/src/components/common/NewTable/Table.styled.ts
+++ b/frontend/src/components/common/NewTable/Table.styled.ts
@@ -154,7 +154,6 @@ export const Table = styled.table(
     word-wrap: break-word;
 
     &:first-of-type {
-      max-width: 530px;
       overflow: hidden;
       text-overflow: ellipsis;
       white-space: nowrap;
--- a/frontend/src/components/common/NewTable/Table.styled.ts
+++ b/frontend/src/components/common/NewTable/Table.styled.ts
@@ -153,24 +153,30 @@ export const Table = styled.table(
     vertical-align: middle;
     word-wrap: break-word;
 
-    & a {
-      color: ${table.td.color.normal};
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/NewTable/Table.styled.ts
```diff
@@ -156,7 +156,7 @@ export const Table = styled.table(
     & a {
       color: ${table.td.color.normal};
       font-weight: 500;
-      max-width: 450px;
+      max-width: 150ch;
```
Haarolean, line 159: as far as I can see we do not use `ch` anywhere yet and not sure how introducing these is legit. @Mgrdich ideas?

Mgrdich: i am not sure about the ch

While the ch unit works as an exact measurement for monospaced / fixed width fonts like Courier, it can be unpredictable with proportional fonts like Arial

Haarolean: @AlexisD02 hey, can you elaborate?

Issue #462, opened 2024-07-02
Consumers: Unsubscribe topics from consumer group


Boniker:
### Issue submitter TODO list

- [X] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md) and the feature is not present there

### Is your proposal related to a problem?

I noticed that the Kafka UI doesn’t have an opportunity to unsubscribe topics from consumer group.

Imagine a situation where you have a lag alert in your environment.
If you reset the offset, the situation will not change because the lag will remain and the topic will be subscribed to this customer group

### Describe the feature you're interested in

I request to be able to completely unsubscribe the topic from the customer group (like remove it from the cg)

AC: 
- stop consumer group
- unsubscribe topics button for all topics
- start consumer group

### Describe alternatives you've considered

we can do it via kafka server (better to do it via Kafka-ui):
```
kafka-consumer-groups --bootstrap-server <server>
   --delete-offsets -group <group_name> \
   --topic <topic_1> \
   --topic <topic_2> \
   --topic <topic_3>
```
### Version you're running

28677a90366791187787f1c7c0c26151648714e3

### Additional context

https://discord.com/channels/897805035122077716/1256173629306048522/1256173629306048522

Boniker:
@Haarolean jfyi

merzenniy:
+1

fienlag:
+1

Haarolean:
Since we have a PR submitted and approved and given the amount of interest in this issue, I'm moving this from 1.2 to an upcoming 1.1 release instead.

PR #549, opened 2024-09-17
Consumers: Unsubscribe topics from consumer group

p-eye:

As mentioned in [issue #462](https://github.com/kafbat/kafka-ui/issues/462), on the consumer group page, add a button to delete topic offsets. It should function similarly to the following command:
```
kafka-consumer-groups --bootstrap-server <server>
   --delete-offsets -group <group_name> \
   --topic <topic_1>
```
image

- [ ] No need to
- [X] Unit checks
- [X] Integration checks


p-eye:
@Haarolean Do you think the review might take a bit longer?
I would appreciate it if your team could review it.
thanks 😃

Haarolean:
> @Haarolean Do you think the review might take a bit longer? I would appreciate it if your team could review it. thanks 😃

it definitely will as I'm trying to wrap the things up for 1.1 release, while the other folks seem to be busy for a review :-)

contract/src/main/resources/swagger/kafbat-ui-api.yaml
```diff
@@ -1048,6 +1048,32 @@ paths:
         200:
           description: OK
 
+  /api/clusters/{clusterName}/consumer-groups/{id}/topics/{topicName}:
+    delete:
+      tags:
+        - Consumer Groups
+      summary: delete consumer group offsets
```
Haarolean, line 1055: In API spec it's 'delete' offsets, in java code it's 'reset' (`.consumerGroupActions(groupId, RESET_OFFSETS)`). I suggest let's decide on one thing. Since we already have 'reset offset' in consumers, I suggest keeping 'reset' in favor of 'delete' here as well. What do you think?

Haarolean (changes requested):
One minor thing regarding delete -> reset naming, the rest (backend) looks good 👍🏼

contract/src/main/resources/swagger/kafbat-ui-api.yaml
p-eye: I named it 'delete' since the kafka-cli uses 'delete-offsets'. 
Also, because this feature deletes the offsets of the consumer group rather than deleting the consumer group itself, I thought the ```RESET_OFFSETS``` was more appropriate than the ```DELETE``` for offset deleting.

And since the "reset consumer group offsets" feature already exists, I think distinguishing it as "delete" would be better. Another option could be splitting the actions, something like ```.consumerGroupActions(groupId, DELETE_OFFSETS)```, just like we do with ```.consumerGroupActions(groupId, RESET_OFFSETS).``` What do you think?

Haarolean: Thanks, these are good points.

Regarding the permission naming, there are a few things I would consider when a question of (not) introducing a new permission arises:
- Consider existing 'sibling' permissions: If we tie new functionality to existing permission, how bad can it be for the end users? What's the chance that somebody wouldn't want to automatically assign permission for such an action to folks who already have the aforementioned permission granted to them? In this particular case, I believe, having permission to reset offsets (to zero) is kind of close to being able to delete offsets altogether, so it seems pretty safe to have a single permission for both actions (and we can always add a new one if we decide otherwise).
- Granularity: We don't want to introduce a new permission unless it's necessary, otherwise it's just an extra overhead with adding more overwhelming choices to grant these new permissions to people.

Given these, I believe using `RESET_OFFSETS` for deleting offsets is okay, what is your opinion on this, @p-eye?

cc @kafbat/backend @germanosin

p-eye: I completely agree with your points !

Haarolean (approved):

Nilumilak (approved):

p-eye:
In addition, I removed unused imports.

germanosin (approved):

Haarolean:
@p-eye thank you for one more contribution!

PR #556, opened 2024-09-21
Brokers: Improve accessibility for r/o clusters

hadisfr:
- [ ] **Breaking change?**

There are some inconsistencies between the backend and frontend in the case of read-only clusters.
This results in enabled buttons which if you click them, get 405 error.
This PR tries to make them consistent.

* Make all configs of brokers read-only for read-only clusters
* Allow "analyze topic" for read-only clusters (https://github.com/provectus/kafka-ui/issues/1967), which seems to do no write-level action on the real data

There may be still some inconsistencies in the ACL page.
I can not check or fix them now, because I have no running kafka cluster with authentication/ACL enabled.

-  No need to
- Unit checks
- Integration checks
- Covered by existing automation

### DIFF
--- a/frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/styled.ts
+++ b/frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/styled.ts
@@ -1,22 +1,23 @@
 import styled from 'styled-components';
 
-export const ValueWrapper = styled.div<{ $isDynamic?: boolean }>`
+export const ValueWrapper = styled.div`
   display: flex;
   justify-content: space-between;
-  font-weight: ${({ $isDynamic }) => ($isDynamic ? 600 : 400)};
+  font-weight: 400;
 
   button {
     margin: 0 10px;
   }
 `;
 
-export const Value = styled.span`
+export const Value = styled.span<{ $isDynamic?: boolean }>`
   line-height: 24px;
   margin-right: 10px;
   text-overflow: ellipsis;
   max-width: 400px;
   overflow: hidden;
   white-space: nowrap;
+  font-weight: ${({ $isDynamic }) => ($isDynamic ? 600 : 400)};
 `;
 
 export const ButtonsWrapper = styled.div`
--- a/api/src/main/java/io/kafbat/ui/config/ReadOnlyModeFilter.java
+++ b/api/src/main/java/io/kafbat/ui/config/ReadOnlyModeFilter.java
@@ -25,7 +25,7 @@ public class ReadOnlyModeFilter implements WebFilter {
       Pattern.compile("/api/clusters/(?<clusterName>[^/]++)");
 
   private static final Set<Pattern> SAFE_ENDPOINTS = Set.of(
-      Pattern.compile("/api/clusters/[^/]+/topics/[^/]+/(smartfilters)$")
+      Pattern.compile("/api/clusters/[^/]+/topics/[^/]+/(smartfilters|analysis)$")
   );
 
   private final ClustersStorage clustersStorage;
--- a/api/src/main/java/io/kafbat/ui/model/InternalBrokerConfig.java
+++ b/api/src/main/java/io/kafbat/ui/model/InternalBrokerConfig.java
@@ -16,12 +16,12 @@ public class InternalBrokerConfig {
   private final boolean isReadOnly;
   private final List<ConfigEntry.ConfigSynonym> synonyms;
 
-  public static InternalBrokerConfig from(ConfigEntry configEntry) {
+  public static InternalBrokerConfig from(ConfigEntry configEntry, boolean readOnlyCluster) {
     InternalBrokerConfig.InternalBrokerConfigBuilder builder = InternalBrokerConfig.builder()
         .name(configEntry.name())
         .value(configEntry.value())
         .source(configEntry.source())
-        .isReadOnly(configEntry.isReadOnly())
+        .isReadOnly(readOnlyCluster || configEntry.isReadOnly())
         .isSensitive(configEntry.isSensitive())
         .synonyms(configEntry.synonyms());
     return builder.build();
--- a/api/src/main/java/io/kafbat/ui/service/BrokerService.java
+++ b/api/src/main/java/io/kafbat/ui/service/BrokerService.java
@@ -59,7 +59,7 @@ public class BrokerService {
     }
     return loadBrokersConfig(cluster, brokerId)
         .map(list -> list.stream()
-            .map(InternalBrokerConfig::from)
+            .map(configEntry -> InternalBrokerConfig.from(configEntry, cluster.isReadOnly()))
             .collect(Collectors.toList()))
         .flatMapMany(Flux::fromIterable);
   }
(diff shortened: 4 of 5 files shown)


api/src/main/java/io/kafbat/ui/controller/BrokersController.java
```diff
@@ -94,10 +97,24 @@ public Mono<ResponseEntity<Flux<BrokerConfigDTO>>> getBrokerConfig(String cluste
         .operationParams(Map.of(BROKER_ID, id))
         .build();
 
+    Flux<BrokerConfigDTO> brokerConfigs = brokerService.getBrokerConfig(getCluster(clusterName), id)
+        .map(clusterMapper::toBrokerConfig);
+
+    var kafkaCluster = clustersStorage.getClusterByName(clusterName)
```
Haarolean, line 103: this is redundant, there's a call `getCluster(clusterName)` on line 100 which can be extracted into a var and reused

api/src/main/java/io/kafbat/ui/controller/BrokersController.java
```diff
@@ -94,10 +97,24 @@ public Mono<ResponseEntity<Flux<BrokerConfigDTO>>> getBrokerConfig(String cluste
...
+        config.setIsReadOnly(true);
+        return config;
+      });
+    }
+
     return validateAccess(context).thenReturn(
         ResponseEntity.ok(
-            brokerService.getBrokerConfig(getCluster(clusterName), id)
-                .map(clusterMapper::toBrokerConfig))
+            brokerConfigs)
```
Haarolean, line 117: we can get the cluster earlier for a var, but let's call getConfigs only after validateAccess, I see two reasons for this:
-  we might not have the permissions to do this (kafka ACL levels) but we'll call the method anyway which may result in unexpected exceptions
- this makes the mono flow more complicated and unreadable with extracted chains

Haarolean (changes requested):

api/src/main/java/io/kafbat/ui/controller/BrokersController.java
```diff
@@ -30,6 +32,7 @@ public class BrokersController extends AbstractController implements BrokersApi
 
   private final BrokerService brokerService;
   private final ClusterMapper clusterMapper;
+  private final Config config;
```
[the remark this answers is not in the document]
hadisfr: fixed in f803510096598a38e7dbbc0ee7ae7dab314cb7f3

api/src/main/java/io/kafbat/ui/controller/BrokersController.java
hadisfr: fixed in f803510096598a38e7dbbc0ee7ae7dab314cb7f3

api/src/main/java/io/kafbat/ui/controller/BrokersController.java
hadisfr: You mentioned an important point. I tried to handle this in 304c880774985d290d412375e196d9ce72fa9015, but I don't like the current state of my code. I would prefer a lambda-like anonymous method instead of `getBrokerConfigIml`. Have you any idea how can I do that?

Haarolean: I belive such a logic shouldn't be a part of the controller. 
I suggest the following:
```patch
Subject: [PATCH] Suggestion
---
Index: api/src/main/java/io/kafbat/ui/service/BrokerService.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/api/src/main/java/io/kafbat/ui/service/BrokerService.java b/api/src/main/java/io/kafbat/ui/service/BrokerService.java
--- a/api/src/main/java/io/kafbat/ui/service/BrokerService.java	(revision 304c880774985d290d412375e196d9ce72fa9015)
+++ b/api/src/main/java/io/kafbat/ui/service/BrokerService.java	(date 1727477557125)
@@ -59,7 +59,7 @@
     }
     return loadBrokersConfig(cluster, brokerId)
         .map(list -> list.stream()
-            .map(InternalBrokerConfig::from)
+            .map(configEntry -> InternalBrokerConfig.from(configEntry, cluster.isReadOnly()))
             .collect(Collectors.toList()))
         .flatMapMany(Flux::fromIterable);
   }
Index: api/src/main/java/io/kafbat/ui/model/InternalBrokerConfig.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/api/src/main/java/io/kafbat/ui/model/InternalBrokerConfig.java b/api/src/main/java/io/kafbat/ui/model/InternalBrokerConfig.java
--- a/api/src/main/java/io/kafbat/ui/model/InternalBrokerConfig.java	(revision 304c880774985d290d412375e196d9ce72fa9015)
+++ b/api/src/main/java/io/kafbat/ui/model/InternalBrokerConfig.java	(date 1727477557130)
@@ -16,12 +16,12 @@
   private final boolean isReadOnly;
   private final List<ConfigEntry.ConfigSynonym> synonyms;
 
-  public static InternalBrokerConfig from(ConfigEntry configEntry) {
+  public static InternalBrokerConfig from(ConfigEntry configEntry, boolean readOnlyCluster) {
     InternalBrokerConfig.InternalBrokerConfigBuilder builder = InternalBrokerConfig.builder()
         .name(configEntry.name())
         .value(configEntry.value())
         .source(configEntry.source())
-        .isReadOnly(configEntry.isReadOnly())
+        .isReadOnly(readOnlyCluster || configEntry.isReadOnly())
         .isSensitive(configEntry.isSensitive())
         .synonyms(configEntry.synonyms());
     return builder.build();

```

What do you think?

hadisfr: Very good idea.

I dropped away my old codes in the controller, applied your suggested changes, and everything seems to work fine.
(I also rebased the branch with main.)

frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/styled.ts
```diff
@@ -1,22 +1,23 @@
 import styled from 'styled-components';
 
-export const ValueWrapper = styled.div<{ $isDynamic?: boolean }>`
+export const ValueWrapper = styled.div`
   display: flex;
   justify-content: space-between;
-  font-weight: ${({ $isDynamic }) => ($isDynamic ? 600 : 400)};
```
Haarolean, line 6: could you explain why we're moving these properties to the value rather than its' wrapper?

api/src/main/java/io/kafbat/ui/controller/BrokersController.java
Haarolean: Thanks for putting a co-authoring for the commit :-)

frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/styled.ts
hadisfr: The previous code made "Property is read-only" bold only for dynamic values in read-only mode, which seems a little weird at that time for me. I expected only the value itself to be bold.
image


BTW, I do not insist in keeping this change.

Haarolean (approved):

Haarolean:
@hadisfr thank you for your first contribution to kafbat UI!

hadisfr:
> @hadisfr thank you for your first contribution to kafbat UI!

Thank you for your help and effort @Haarolean!

Issue #183, opened 2024-03-01
KC: Support First-class offsets in Kafka Connect


Haarolean:
From https://github.com/provectus/kafka-ui/issues/4289

 ### Describe the feature you're interested in
 [KIP-875](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect) just debuted on Kafka 3.6 and it would be great to be able to leverage this feature using the UI
 
 Provides first-class admin support for offsets in Kafka Connect. KIP-875 Part 1 added endpoints to get offsets and a new STOPPED state for connectors. The alter offsets and reset offsets endpoints have now been added.
 
| Action | Description |
----------|----------|
|  GET /connectors/{connector}/offsets |	Retrieve the offsets for a connector; the connector must exist |
| PATCH /connectors/{connector}/offsets |	Alter the offsets for a connector; the connector must exist, and must be in the STOPPED state |
| DELETE /connectors/{connector}/offsets	| Reset the offsets for a connector; the connector must exist, and must be in the STOPPED state |
| PUT /connectors/{connector}/pause	| Pause the connector; the connector must exist |

yeikel:
- When the "Reset Offset" button is used there is no visual feedback so it seems that the UI did not do anything when it actually did in the backend  (as per the network activity)

Haarolean:
@yeikel could you please raise a bug if it's still valid?

PR #573, opened 2024-10-04
KC: Stop Connectors and Reset Connector Offsets

Dugong42:

This brings a few features to the "Connector" pages (list and detail), related to Connector Offsets management (#183)

### Added
- Added a dropdown option to "Stop" a connector
- Added a dropdown option to "Reset Connector Offsets". It requires a confirmation as for the "Delete" action.
- Added a new permission "RESET_OFFSETS" for connect
- Displayed the "Pause" and "Resume" actions in the Connectors list page dropdown (previously only on the connector detail page)

### Changes
- The "Restart" menu of the detail page is disabled during connector state changes, instead of its individual items being disable, to give a better visual feedback of the ongoing operation.
- The "Remove Connector" action on the connector list page is renamed as "Delete" to match the same action on the detail page and other similar actions on other resources.

### Fixed
- _edit2:_ On the detail page, Connector actions now behave as expected, triggering a data refresh

### Misc
Given I'm working behind a corporate proxy which was a pain to configure, I added a few tweaks to improve the dev experience.
- Added some configuration to be able to build behind a corporate proxy with minimal config (for instance proxies/mirrors in .npmrc and .m2/settings.xml and system proxies) 
- Reduced maven dependency fetch time: Moved the declaration of the more specific Confluent repository after the more common Apache repo. This matters in my experience because maven tries them in order for every package.

_edit1 with screenshots_
> The "Reset" button is disabled if the connector is not STOPPED
image

> On the detail page, the "Restart" dropdown now contains the new "Stop" action
image

> The other dropdown contains the Reset and Delete actions
image

> Confirmation dialog
image

- Not too sure about the UX of the "Restart" dropdown. It's a bit weird to have part of the actions in it and part in the other dropdown.
- Not sure about the current unit test policy, did not see many tests for other similar features.
- ~VSCode applied some formatting changes. I hope that's not a bother to review.~
- I tested manually but was unable to run java unit tests locally, I have an 8GB RAM toaster which apparently is not enough to run WSL + the tests, as they kept getting killed by OOM errors.

- [ ] No need to
- [x] Unit checks
- [x] Integration checks

  - Frontend tests are passing
  - Backend tests kept crashing because of out of memory errors. My toaster with 8GB RAM was not enough to run WSL + the tests apparently.
  - Needs some documentation for the new permission

baby_kapibara


contract/src/main/resources/swagger/kafbat-ui-api.yaml
```diff
@@ -2527,7 +2542,7 @@ components:
         partitions:
           type: array
           items:
-            $ref: "#/components/schemas/Partition"
```
Haarolean, line 2530: Could we revert all unnecessary formatting changes? Makes reviewing the PR complicated. Happy to discuss the necessity of formatting changes in a separate PR.

Haarolean (changes requested):

contract/src/main/resources/swagger/kafbat-ui-api.yaml
Dugong42: I rolled back the formatting changes :)

Dugong42:
I fixed the data-refresh during connectors state changes (see description)

Dugong42:
There seems to be a delay between the Connect API response to the `STOP` operation and the connector actually behind in the `STOPPED` state. 

Since the `GET` to refresh the state is done immediately after this operation, it often leads to the displayed state remaining shown as `RUNNING`. Any subsequent refresh shows the correct state.

api/src/main/java/io/kafbat/ui/controller/KafkaConnectController.java
```diff
@@ -285,4 +285,26 @@ private Comparator<FullConnectorInfoDTO> getConnectorsComparator(ConnectorColumn
       default -> defaultComparator;
     };
   }
+
+  @Override
+  public Mono<ResponseEntity<Void>> resetConnectorOffsets(String clusterName, String connectName,
```
Haarolean, line 290: Can we add some tests for the new functionality? 
`KafkaConnectServiceTests` looks like a good place to start

api/src/main/java/io/kafbat/ui/controller/KafkaConnectController.java
```diff
@@ -285,4 +285,26 @@ private Comparator<FullConnectorInfoDTO> getConnectorsComparator(ConnectorColumn
...
+      String connectorName,
+      ServerWebExchange exchange) {
+    ConnectAction[] connectActions;
+
+    connectActions = new ConnectAction[] { ConnectAction.VIEW, ConnectAction.RESET_OFFSETS };
+
+    var context = AccessContext.builder()
+        .cluster(clusterName)
+        .connectActions(connectName, connectActions)
```
Haarolean, line 299: more of a code style preference, imo it'd look better as a list of static imports:
`.connectActions(connectName, VIEW, RESET_OFFSETS)`, what do you think?

```suggestion
        .connectActions(connectName, VIEW, RESET_OFFSETS)
```

Haarolean (approved):
The backend looks good! Would be great to have some new tests as well

api/src/main/java/io/kafbat/ui/controller/KafkaConnectController.java
Dugong42: I agree, that's much better. I'll do the change.

yeikel:
Ideally, we should also add the `PUT /connectors/connector/resume ` option to resume a connector after it has been stopped

api/src/main/java/io/kafbat/ui/controller/KafkaConnectController.java
Dugong42: Sure, I'll look into that

Dugong42:
I've added a couple of tests for the reset connectors feature. 

I've had to update the images used in the tests to confluent `7.7.1` which is the current latest version, because this feature requires the kafka connect `/connectors/{}/offsets` API introduced in `7.6` (kafka 3.6) by [KIP-875](https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect)

The image upgrade introduced a bug in the tests because some kafka connect errors that were previously processed asynchronously are now rising up to the `POST|PUT /connectors` responses. See the [KAFKA-14455](https://issues.apache.org/jira/browse/KAFKA-14455) issue.

Specifically it now randomly produces HTTP 500 errors because Kafka Connect often falls into the "rebalancing" state when we create/delete connectors in a row.

My workaround is to optimistically ignore the connector creation error and then check for its existence.

yeikel:
@Dugong42 FYI: https://github.com/kafbat/kafka-ui/pull/710#discussion_r1889421217

fallen-up:
tested and
1) it's strange that acl doesn't fit into a unified formula.
now you need to do an acl like this (uppercase):
```
          - resource: connect
            value: ".*"
            actions: ["view", "restart", "RESET_OFFSETS"]
```
2) After stopping connector - Error 500
`JSON decoding error: Cannot construct instance of `io.kafbat.ui.connect.model.ConnectorStatusConnector$StateEnum`, problem: Unexpected value 'STOPPED'`

Haarolean:
> 1. now you need to do an acl like this (uppercase):

@fallen-up this shouldn't be a case, it's case insensitive for all the enums. Can you confirm please?

Haarolean:
@Dugong42 hi, can you take a look at the 2nd issue @fallen-up mentioned? And I guess we don't need the workaround for the KC anymore considering @yeikel's changes

fallen-up:
> @fallen-up this shouldn't be a case, it's case insensitive for all the enums. Can you confirm please?
@Haarolean yep, looks like my bad. register is insensitive

Dugong42:
@yeikel thanks for the proper fix!

@Haarolean I've rolled back my workaround and taken the retry mechanic into account. 

I've tested different values for the `actions` and could not replicate the case issue.
As @fallen-up said in the end it seems to work as expected.

It should be ready to merge :)

pom.xml
```diff
@@ -73,10 +73,6 @@
     </properties>
 
     <repositories>
-        <repository>
-            <id>confluent</id>
-            <url>https://packages.confluent.io/maven/</url>
```
yeikel, line 78: What's the purpose of this change? It seems unrelated

Dugong42: Kinda unrelated yes, I got annoyed during development of the long package loading time on a slow connexion.

This quickens the fetching of packages as maven checks the repos in order, so it's preferable to move the more specific confluent repo last.

I also brought a few changes to the default maven/pnpm settings to help developing behind a corporate proxy.

frontend/src/components/Connect/List/ActionsCell.tsx
```diff
@@ -58,9 +64,25 @@ const ActionsCell: React.FC<CellContext<FullConnectorInfo, unknown>> = ({
...
+
+  const stopConnectorHandler = () =>
+    stateMutation.mutateAsync(ConnectorAction.STOP);
+
+  const resetOffsetsHandler = () => {
+    confirm(
+      <>
+        Are you sure you want to reset the <b>{name}</b> connector offsets?
+      </>,
+      () => resetConnectorOffsetsMutation.mutateAsync()
```
Mgrdich, line 78: one question just for curiosity why are we using `mutateAsync`  instead of `mutate` does the confirm wait for the promise to be resolved

pom.xml
Dugong42: @Haarolean Do you think that requires a separate MR ?

frontend/src/components/Connect/List/ActionsCell.tsx
Dugong42: Although you raise a good point because it don't think it does await connector state changes (`PAUSE`, `STOP`...). 
In those cases the UI does not always display the new state... but that's another story.

pom.xml
Haarolean: @Dugong42 not really, let's keep it here.

Haarolean:
@Dugong42 could you please refrain from force-pushing a PR branch that has been previously reviewed in the future? This marks multiple files as unreviewed without a clear diff even if there are none since the last review.

Haarolean (approved):

yeikel (commented):
Thank you! (I can only comment as my approval does not count)

Haarolean:
> Thank you! (I can only comment as my approval does not count)

yep, this is a limitation to prevent random users LGTMing on PRs (yes, it's actually a thing)

yeikel (approved):

Haarolean:
@Dugong42 thanks for your first contribution to kafbat UI! 🎉

fallen-up:
finally managed to do a full test.
@Dugong42 
it seems that the initial description of the PR, including the names of the buttons and other elements, was incorrect. here are my reasons:
- "Reset offset" suggests resetting to a specific value. In other functionalities of kafka-ui, "reset" clearly means triggering a reset of one type (earliest, latest, timestamp, offset).
- there is also an option to delete a consumer group.
- in https://github.com/kafbat/kafka-ui/pull/636, the deletion of offsets for a single topic within a group is implemented, and in the menu, it appears as "Delete offsets".
- this PR simply performs a **Delete**, not Patch/Reset. it can easily lead to confusion, and at a minimum, the button labels should be changed.

fallen-up:
> Is the challenge that the name is misleading?
> 
> The original API also uses the "delete" terminology, while what it does is to "reset it". For the context of offset tracking delete and reset seem to be the same thing here

let's go the other way around.
suppose there is a kind person who implements "PATCH /connectors/{connector}/offsets" - how should he call the button? "patch"?
if so, why can't the current one be called "delete"? and then you won't need to look at the documentation to understand what it does.

PR #579, opened 2024-10-06
UX: Messages: Use "Newest" as default mode

### LINKED ISSUE #570: UX: Messages: Use "Newest" as default mode
### Issue submitter TODO list

Most of the time, we have experienced that users want to see the newest first

In here:

https://github.com/kafbat/kafka-ui/blob/3f3ca4cfabe36acd57fddcec1589c39a1800f61f/frontend/src/lib/hooks/useMessagesFilters.ts#L38

Use default 1, which would be `latest`

Making it a configuration parameter, but I think "Newest" is good enough. I'd like to check here if everyone also agrees that Newest as default would be the best default. 

### Version you're running

1.0.0

### Additional context

_No response_

polegkashti:

Simply chaned the default filter mode.

Hi, I’ve noticed that other users have also expressed a need for this change, and some of our Kafka users specifically mentioned it. I believe it’s a great idea to modify the default mode, as it aligns better with Kafka's real-time streaming nature.

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

gamer cat

### DIFF
--- a/frontend/src/lib/hooks/filterUtils.ts
+++ b/frontend/src/lib/hooks/filterUtils.ts
@@ -1,8 +1,8 @@
 import { PollingMode } from 'generated-sources';
 
 export const ModeOptions = [
-  { value: PollingMode.EARLIEST, label: 'Oldest' },
   { value: PollingMode.LATEST, label: 'Newest' },
+  { value: PollingMode.EARLIEST, label: 'Oldest' },
   { value: PollingMode.TAILING, label: 'Live' },
   { value: PollingMode.FROM_OFFSET, label: 'From offset' },
   { value: PollingMode.TO_OFFSET, label: 'To offset' },


polegkashti:
resolves #570

Haarolean (approved):

fallen-up:
i tried to solve the problem this way. as a result, sometimes "Newest" is displayed in the UI, but the messages are empty. only changing to "Oldest" and then back to "Newest" helps.

Haarolean:
> i tried to solve the problem this way. as a result "Newest" is displayed in the UI, but sometimes the messages are empty. only changing to "Oldest" and then back to "Newest" helps.

should be working, as we have `const defaultModeValue = ModeOptions[0].value;` in `useMEssageFilters.ts`. @polegkashti can you confirm?

polegkashti:
> > i tried to solve the problem this way. as a result "Newest" is displayed in the UI, but sometimes the messages are empty. only changing to "Oldest" and then back to "Newest" helps.
> 
> should be working, as we have `const defaultModeValue = ModeOptions[0].value;` in `useMEssageFilters.ts`. @polegkashti can you confirm?

Based on my testing, everything is functioning properly, and I don't see any reason for this issue to occur.

polegkashti:
I tested the change again with a variety of messages and was unable to reproduce the issue. Additionally, I reviewed the codebase for any other references to the "Newest" filter and found no related components.

Given this, I believe there isn't a valid reason to withhold approval for this pull request. If the issue continues to affect you, I recommend opening a new issue with more details to help us investigate further :)

Haarolean:
@polegkashti thank you!

Issue #26, opened 2024-01-23
Online partition count is red


Haarolean:
https://discord.com/channels/897805035122077716/1197881542052159608

![IMG_2588](https://private-user-images.githubusercontent.com/59020701/304126526-139d1442-0dc8-4727-b45f-73ba1b21f320.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDc3NTAxMTcsIm5iZiI6MTcwNzc0OTgxNywicGF0aCI6Ii81OTAyMDcwMS8zMDQxMjY1MjYtMTM5ZDE0NDItMGRjOC00NzI3LWI0NWYtNzNiYTFiMjFmMzIwLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjEyVDE0NTY1N1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTljOGYzNmRhNGE4YTI1NTIzZmE4ZTg1ZGQxNWI5MzZlZTA2M2NiYjQxYmZmNDM1ODc0YzJjY2EyM2MwYzE0NTgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.KRKKwoS1mkUQMP0SO4vO_AUBSst3GfkZHClzvk45Zww)

brokers.json:
```
[
    {
        "id": 1,
        "host": "test-region-1-broker-1.kafka",
        "port": 9092,
        "bytesInPerSec": null,
        "bytesOutPerSec": null,
        "partitionsLeader": 18,
        "partitions": 110,
        "inSyncPartitions": 110,
        "partitionsSkew": 0.0,
        "leadersSkew": -1.8
    },
    {
        "id": 2,
[...]

PR #607, opened 2024-10-13
FE: Fix 'Online Partitions' column

### LINKED ISSUE #526: Brokers: Wrong 'Online partition count' value
### Issue submitter TODO list

- [X] I've looked up my issue in [FAQ](https://ui.docs.kafbat.io/faq/common-problems)
- [X] I've tried running `main`-labeled docker image and the issue still persists there
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md)

Kafka UI used to show the actual partition count per broker in the "Online Partitions" column under "Brokers" tab. After upgrading to v1.0.0, each broker row shows the total number of online partitions instead. Weirdly, the "partition skew" calculations are still accurate.

For example take a balanced cluster with 178 partitions, rep factor of 3, and 4 brokers. There are a total of 534 replicas in the cluster and each broker row in the table should show 133 or 134 online partitions. However, all of the rows are showing 178. 

### Expected behavior

Each broker should show the correct number of online partition replicas it holds. Given the example above, it should be 133 or 134 per broker.

### Your installation details

I'm using v1.0.0 with a generic config file, nothing unique.

### Steps to reproduce

Deploy kafka ui and link it to any cluster.  I've tried it with 4 different clusters and they all show the same thing.

### Screenshots

_No response_

### Logs

_No response_

### Additional context

_No response_

bachmanity1:

Resolves #526.

The value shown in the `Online Partitions` column of the brokers list table was modified in #137. This PR restores the original logic used to render `Online Partitions` prior to #137.

- [ ] No need to
- [ ] Unit checks
- [x] Integration checks

### DIFF
--- a/e2e-tests/src/main/java/io/kafbat/ui/screens/brokers/BrokersList.java
+++ b/e2e-tests/src/main/java/io/kafbat/ui/screens/brokers/BrokersList.java
@@ -48,8 +48,8 @@ public class BrokersList extends BasePage {
   }
 
   private List<SelenideElement> getEnabledColumnHeaders() {
-    return Stream.of("Broker ID", "Disk usage", "Partitions skew",
-            "Leaders", "Leader skew", "Online partitions", "Port", "Host")
+    return Stream.of("Broker ID", "Disk usage", "In Sync Replicas", "Replicas", "Replicas skew",
+            "Leaders", "Leaders skew", "Port", "Host")
         .map(name -> $x(String.format(columnHeaderLocator, name)))
         .collect(Collectors.toList());
   }
--- a/frontend/src/components/Brokers/BrokersList/SkewHeader/SkewHeader.tsx
+++ b/frontend/src/components/Brokers/BrokersList/SkewHeader/SkewHeader.tsx
@@ -6,7 +6,7 @@ import * as S from './SkewHeader.styled';
 
 const SkewHeader: React.FC = () => (
   <S.CellWrapper>
-    Partitions skew
+    Replicas skew
     <Tooltip
       value={<InfoIcon />}
       content="The divergence from the average brokers' value"
--- a/frontend/src/components/Brokers/BrokersList/TableCells/TableCells.tsx
+++ b/frontend/src/components/Brokers/BrokersList/TableCells/TableCells.tsx
@@ -78,21 +78,12 @@ type InSyncReplicasProps = CellContext<
   BrokersTableRow['inSyncReplicas']
 >;
 
-export const InSyncReplicas = ({ row }: InSyncReplicasProps) => {
-  const { replicas, inSyncReplicas } = row.original;
+export const InSyncReplicas = ({ getValue }: InSyncReplicasProps) => {
+  return <ColoredCell value={getValue() || ''} />;
+};
 
-  if (
-    replicas === undefined ||
-    inSyncReplicas === undefined
-  ) {
-    return null;
-  }
-  const value = `${inSyncReplicas} of ${replicas}`;
+type ReplicasProps = CellContext<BrokersTableRow, BrokersTableRow['replicas']>;
 
-  return (
-    <ColoredCell
-      value={value}
-      attention={inSyncReplicas < replicas}
-    />
-  );
+export const Replicas = ({ getValue }: ReplicasProps) => {
+  return <ColoredCell value={getValue() || ''} />;
 };
--- a/frontend/src/components/Brokers/BrokersList/TableCells/TableCells.tsx
+++ b/frontend/src/components/Brokers/BrokersList/TableCells/TableCells.tsx
@@ -57,7 +57,7 @@ export const DiscUsage = ({
 
 type ScewProps = CellContext<
   BrokersTableRow,
-  BrokersTableRow['partitionsSkew'] | BrokersTableRow['leadersSkew']
+  BrokersTableRow['replicasSkew'] | BrokersTableRow['leadersSkew']
 >;
 
 export const Skew = ({ getValue }: ScewProps) => {
@@ -73,25 +73,26 @@ export const Skew = ({ getValue }: ScewProps) => {
   );
 };
 
-type OnlinePartitionsProps = CellContext<
+type InSyncReplicasProps = CellContext<
   BrokersTableRow,
-  BrokersTableRow['onlinePartitionCount']
+  BrokersTableRow['inSyncReplicas']
 >;
 
-export const OnlinePartitions = ({ row }: OnlinePartitionsProps) => {
-  const { onlinePartitionCount, offlinePartitionCount } = row.original;
+export const InSyncReplicas = ({ row }: InSyncReplicasProps) => {
+  const { replicas, inSyncReplicas } = row.original;
 
   if (
-    onlinePartitionCount === undefined ||
-    offlinePartitionCount === undefined
+    replicas === undefined ||
+    inSyncReplicas === undefined
   ) {
     return null;
   }
+  const value = `${inSyncReplicas} of ${replicas}`;
 
   return (
     <ColoredCell
-      value={onlinePartitionCount}
-      attention={offlinePartitionCount > 0}
+      value={value}
+      attention={inSyncReplicas < replicas}
     />
   );
 };
--- a/frontend/src/components/Brokers/BrokersList/TableCells/TableCells.tsx
+++ b/frontend/src/components/Brokers/BrokersList/TableCells/TableCells.tsx
@@ -78,8 +78,15 @@ type InSyncReplicasProps = CellContext<
   BrokersTableRow['inSyncReplicas']
 >;
 
-export const InSyncReplicas = ({ getValue }: InSyncReplicasProps) => {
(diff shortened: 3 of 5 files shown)


Haarolean:
>This PR restores the original logic 
@bachmanity1 hi, could you please elaborate here? The changes previously introduced were done so with a purpose (even if introducing incorrect behavior), so reverting these changes you're probably introducing another faulty behavior which was present before #137 (#26)?

bachmanity1:
> > This PR restores the original logic
> 
> @bachmanity1 hi, could you please elaborate here? The changes previously introduced were done so with a purpose (even if introducing incorrect behavior), so reverting these changes you're probably introducing another faulty behavior that was present before #137 (#26)?

Hi @Haarolean, thanks for review! 

You're right. We can't simply revert the changes made in #137 because that would bring back the original faulty behavior. While the changes in #137 fixed the issue with the `Online Partitions` column's color, they also caused every row in the brokers table to show the same value in that column. From looking at the backend code, it seems that `Online Partitions` refers to partitions where the `describeTopics` API doesn't return null for the partition leader. While it's possible to compute this for the whole cluster, I don't think it's compute to do so for each individual broker.

So, how about renaming the `Online Partitions` column to just `Partitions` and showing the actual number of partitions (replicas) on each broker? If possible, I'd even suggest renaming `Online Partitions` to `Replicas` and `Partitions Skew` to `Replicas Skew`. I think this would make more sense, since in Kafka terminology, partitions stored on brokers are referred to as replicas.

bachmanity1:
Hi @Haarolean,

Following my last comment I've slightly changed brokers table by rearranging order and renaming columns. Here is sample screenshot:

image

bachmanity1:
Hi @Haarolean, can you have a look at this?

bachmanity1:
Hi @Haarolean, can you have a look at this?

Haarolean:
> While it's possible to compute this for the whole cluster, I don't think it's possible (or needed?) to do so for each individual broker.

Why 'not needed'? As a user, I'd like to know which broker has the problem.

> So, how about renaming the `Online Partitions` column to just `Partitions`

These are two different things, partitions are a sum of online + offline partitions. Current implementation returns these as two distinct fields (`onlinePartitionCount` and `offlinePartitionCount`) and masquerading one for another on the frontend doesn't look good.

> If possible, I'd even suggest renaming `Online Partitions` to `Replicas` and `Partitions Skew` to `Replicas Skew`. I think this would make more sense, since in Kafka terminology, partitions stored on brokers are referred to as replicas.

I don't think it's a good idea as this conflates the terms partition and replica. While a replica is a physical copy of a partition stored on a broker, at the same time a partition can have **multiple** replicas (one lead and multiple followers) spread across different brokers.

@bachmanity1

bachmanity1:
> I don't think it's a good idea as this conflates the terms partition and replica. While a replica is a physical copy of a partition stored on a broker, at the same time a partition can have multiple replicas (one lead and multiple followers) spread across different brokers.

This is exactly why I suggested using the term `replica` instead of `partition`. I'm just a bit confused about how we can say that a particular broker stores a certain number of partitions or that a partition stored on a broker is online/offline when the term partition is not a physical thing but rather a logical concept (like a topic) that is distributed across multiple brokers. I’ve never seen the term `partition` used this way in any Kafka literature, but please let me know if I’m mistaken.

Haarolean:
> > I don't think it's a good idea as this conflates the terms partition and replica. While a replica is a physical copy of a partition stored on a broker, at the same time a partition can have multiple replicas (one lead and multiple followers) spread across different brokers.
> 
> This is exactly why I suggested using the term `replica` instead of `partition`. I'm just a bit confused about how we can say that a particular broker stores a certain number of partitions or that a partition stored on a broker is online/offline when the term partition is not a physical thing but rather a logical concept (like a topic) that is distributed across multiple brokers. I’ve never seen the term `partition` used this way in any Kafka literature, but please let me know if I’m mistaken.

Ah, this makes sense indeed in the context of displaying data for a broker. The problem is that we can't just rename the columns, as they represent different data (partitions vs replicas). This might require backend changes if the desired data is not yet exposed in a format acceptable to us.

bachmanity1:
> The problem is that we can't just rename the columns, as they represent different data (partitions vs replicas). This might require backend changes if the desired data is not yet exposed in a format acceptable to us.

I believe all the necessary data is already available; it just seems like there's some confusion around naming.

Looking at the backend code here: https://github.com/kafbat/kafka-ui/blob/582e9a9902cb3114d079a2d9ad0a5ed495ceadb2/api/src/main/java/io/kafbat/ui/model/PartitionDistributionStats.java#L39-L49

and this contract https://github.com/kafbat/kafka-ui/blob/582e9a9902cb3114d079a2d9ad0a5ed495ceadb2/contract/src/main/resources/swagger/kafbat-ui-api.yaml#L2843-L2865

It seems the `partitions` field represents the total number of replicas, the `inSyncPartitions` field represents in-sync replicas, and the `partitionsSkew` field represents replica skew. So, I think it's fine to rename the columns in the FE as long as they use the correct data from the BE. We could rename the BE APIs in a separate PR or include it in this one.

bachmanity1:
@Haarolean If you don't have any objections, how about merging this PR? We can rename columns in the separate PR?

bachmanity1:
Hi @Haarolean, can we at least decide what to do with the `Online Partitions`? Currently, there is clearly a problem with it.

Haarolean:
It seems to display proper info we need to fix the following:

1. Column `Online partitions`:
- rename to `In sync partitions`
- Display value from `/api/clusters/local/brokers` endpoint, field `inSyncPartitions`

2. Add column `partitions`:
- display value from `/api/clusters/local/brokers` endpoint, field `partitions`

3. Show red values in `In sync partitions` if `inSyncPartitions` value is less than `partitions` for this broker

@bachmanity1 is that right?

bachmanity1:
@Haarolean 

> @bachmanity1 is that right?

Yes, looks right. I would just suggest renaming `partitions` to `replicas`, so columns would change like:

1. Rename `Online Partitions` to `In Sync Replicas`
     - display value from `/api/clusters/local/brokers` endpoint, field `inSyncPartitions`
2. Add new column `Replicas`
    - display value from `/api/clusters/local/brokers` endpoint, field `partitions`
3.  Rename `Partitions skew` to `Replicas skew`
    - display value from `/api/clusters/local/brokers` endpoint, field `partitionsSkew`

germanosin:
Agree on ISR, but let's retain the names for Partitions and Partitions skew.

> @Haarolean
> 
> > @bachmanity1 is that right?
> 
> Yes, looks right. I would just suggest renaming `partitions` to `replicas`, so columns would change like:
> 
> 1. Rename `Online Partitions` to `In Sync Replicas`
>    
>    * display value from `/api/clusters/local/brokers` endpoint, field `inSyncPartitions`
> 2. Add new column `Replicas`
>    
>    * display value from `/api/clusters/local/brokers` endpoint, field `partitions`
> 3. Rename `Partitions skew` to `Replicas skew`
>    
>    * display value from `/api/clusters/local/brokers` endpoint, field `partitionsSkew`

Haarolean:
> > I don't think it's a good idea as this conflates the terms partition and replica. While a replica is a physical copy of a partition stored on a broker, at the same time a partition can have multiple replicas (one lead and multiple followers) spread across different brokers.
> 
> This is exactly why I suggested using the term `replica` instead of `partition`. I'm just a bit confused about how we can say that a particular broker stores a certain number of partitions or that a partition stored on a broker is online/offline when the term partition is not a physical thing but rather a logical concept (like a topic) that is distributed across multiple brokers. I’ve never seen the term `partition` used this way in any Kafka literature, but please let me know if I’m mistaken.

@germanosin what's your take on this?

Haarolean:
@bachmanity1 would you like to update the PR with the changes we discussed?

bachmanity1:
@germanosin 
> Agree on ISR, but let's retain the names for Partitions and Partitions skew.

I think naming replicas as partitions can be confusing for users. In the image below, the **Partitions** section (above the brokers table) shows that 64 out of 64 partitions are online. This makes me think there are 64 partitions in my cluster (which is true). However, when I look at the brokers table and sum up the **Partitions** column for each broker (64 + 52), the total is much larger than 64. This is confusing because the numbers don’t match. I believe renaming the **Partitions** column to **Replicas** would resolve this confusion.

image

germanosin:
@bachmanity1, thanks for the explanation. I agree—let's go with Replicas as the name.

Haarolean:
@bachmanity1 hey, would you like to update your PR?

bachmanity1:
> @bachmanity1 hey, would you like to update your PR?

Sorry, I was a little busy lately. I'm gonna update this PR in a couple of days.

germanosin:
> > @bachmanity1 hey, would you like to update your PR?
> 
> Sorry, I was a little busy lately. I'm gonna update this PR in a couple of days.
No worries at all—take your time, there’s no rush. And once again, thank you for your contribution!

germanosin (approved):

Haarolean (approved):

PR #608, opened 2024-10-13
feat(broker-configs): adjust edit button state in read-only mode

### LINKED ISSUE #230: FE: Adjust button states w/ r/o mode
There is a cluster read-only mode. Not all the altering actions on the frontend are unavailable when a cluster is r/o.
Fix this, perhaps add a check on button/other component level.

spartanns:

Closes #230 

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

### DIFF
--- a/frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/index.tsx
+++ b/frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/index.tsx
@@ -24,7 +24,6 @@ const InputCell: FC<InputCellProps> = ({ row, onUpdate }) => {
     source,
     value: initialValue,
     isSensitive,
-    isReadOnly,
   } = row.original;
 
   const handleSave = (newValue: string) => {
@@ -52,7 +51,6 @@ const InputCell: FC<InputCellProps> = ({ row, onUpdate }) => {
       onEdit={() => setIsEdit(true)}
       isDynamic={isDynamic}
       isSensitive={isSensitive}
-      isReadOnly={isReadOnly}
     />
   );
 };
--- a/frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/InputCellViewMode.tsx
+++ b/frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/InputCellViewMode.tsx
@@ -1,4 +1,4 @@
-import React, { type FC } from 'react';
+import React, { type FC, useContext } from 'react';
 import { Button } from 'components/common/Button/Button';
 import EditIcon from 'components/common/Icons/EditIcon';
 import type { ConfigUnit } from 'components/Brokers/Broker/Configs/lib/types';
@@ -6,6 +6,7 @@ import Tooltip from 'components/common/Tooltip/Tooltip';
 import { getConfigDisplayValue } from 'components/Brokers/Broker/Configs/lib/utils';
 
 import * as S from './styled';
+import ClusterContext from 'components/contexts/ClusterContext';
 
 interface InputCellViewModeProps {
   value: string;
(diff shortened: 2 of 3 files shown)


Haarolean:
@spartanns there have been multiple inconsistencies across the app regarding the r/o mode, is this all the changes we need?

spartanns:
Basically, if a cluster is read only, then ALL the broker config edit functionality is disabled. Meaning nothing in broker config can be edited in cluster RO mode.

Haarolean:
there have been issues across the app as I mentioned in #230, so the purpose of that issue is to fix these discrepancies across the whole app

spartanns:
I tested my changes with 2 clusters in the app. One was read only the other one was not... You can take a look in this video:
[Untitled Video October 16, 2024 11_18 PM.webm](https://github.com/user-attachments/assets/568f287d-5da2-4b69-a5f4-138ca02fb833)

In case I missed something, let me know and I'll fix it.

Haarolean:
Once again -- the issue is not *just* about broker configs, it's about all the possible buttons in all the places.
For instance, in a r/o cluster restart/remove connectors options are available (they shouldn't be):
image
We have to check for all these things.

spartanns:
Ahhh, okay, I didn't know that...I'll get right on it, and I'll let you know once I'm done.

spartanns:
Okay, so I disabled the ActionsCell in connectors table in cluster RO mode:
Screenshot from 2024-10-17 15-10-12
Broker config edits are disabled as well as all topic configs. Connector creation is also disabled in RO mode.

If there's anything else, just let me know ;p

frontend/src/components/Connect/List/ActionsCell.tsx
```diff
@@ -59,7 +61,7 @@ const ActionsCell: React.FC<CellContext<FullConnectorInfo, unknown>> = ({
     stateMutation.mutateAsync(ConnectorAction.RESTART_FAILED_TASKS);
 
   return (
-    <Dropdown>
+    <Dropdown disabled={isReadOnly}>
```
Haarolean, line 64: Not sure it's a good idea to apply this kind of rule for the whole component, as some elements in the dropdown might be allowed (like, r/o actions). There are also exclusions from r/o rule - see `SAFE_ENDPOINTS` in `io/kafbat/ui/config/ReadOnlyModeFilter.java:23`:
```
private static final Set<Pattern> SAFE_ENDPOINTS = Set.of(
      Pattern.compile("/api/clusters/[^/]+/topics/[^/]+/(smartfilters|analysis)$")
  );
```
So, actions related to topic analysis and smartfilters should be allowed despite being non-GET HTTP methods.

Haarolean (changes requested):

frontend/src/components/Connect/List/ActionsCell.tsx
spartanns: Fixed. Applied additional isReadOnly check to only restart/remove connector actions.

frontend/src/components/Brokers/Broker/Configs/TableComponents/InputCell/index.tsx
```diff
@@ -38,6 +33,7 @@ const InputCell: FC<InputCellProps> = ({ row, onUpdate }) => {
 
   const isDynamic = source === 'DYNAMIC_BROKER_CONFIG';
   const configUnit = getConfigUnit(name);
+  const { isReadOnly } = useContext(ClusterContext);
```
Haarolean, line 36: we shouldn't neglect the `isReadOnly` property of the row, it has a different reason to be disabled -- a broker config property might be r/o

Haarolean (changes requested):

Haarolean:
@spartanns thanks for your first PR! Sorry for the delay.

PR #669, opened 2024-11-22
Resolves kafbat/kafka-ui#90 Added contract profile, to only build API, local profile now builds frontend

### LINKED ISSUE #90: Refactor maven profiles
Current issues:
1. Some users need an API build without a frontend. There's a "skipuibuild" property for that. Doesn't seem optimal.
2. Local build doesn't build frontend, which is usually required for a local run. Prod profile, on the other hand, builds a docker image, no way to easily disable this. Which is usually not required for a local run.

KliminV:
Added contract profil, to only build API
Local profile now builds frontend

Resolves #90

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

- [x ] I have performed a self-review of my own code


KliminV:
I think no unit tests are needed. However I would be happy to update documentation if needed

KliminV:
@Haarolean could you please take a look?

pom.xml
```diff
@@ -193,4 +187,25 @@
         </pluginManagement>
     </build>
 
+    <profiles>
+        <profile>
+            <id>all</id>
```
Haarolean, line 192: why would we need this `all` profile?

pom.xml
```diff
@@ -193,4 +187,25 @@
...
+            </activation>
+            <modules>
+                <module>api</module>
+                <module>contract</module>
+                <module>e2e-tests</module>
+                <module>serde-api</module>
+            </modules>
+        </profile>
+        <profile>
+            <id>contract</id>
```
Haarolean, line 204: We don't quite need the `contract` profile. To achieve the same one can build just one module, `contract`.
What we need is a jar without the frontend, basically what skipUiBuild does now, but as a profile.

api/pom.xml
```diff
@@ -399,12 +400,64 @@
...
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
     <profiles>
+        <profile>
+            <id>local</id>
+            <properties>
+                <skipUIBuild>false</skipUIBuild>
```
Haarolean, line 453: We'd need to get rid of this property altogether replacing it with a separate profile for transparency

Haarolean (changes requested):

pom.xml
KliminV: That's default - to build all modules if no profile provided.

Let's confirm scope:
- we need to get rid of skipUiBuild and replace it with profile
- add frontend as part of local build

Is that all? 

Then I suggest declare frontend as separate module - that allows us to avoid using properties to skip build stages

pom.xml
KliminV: should be solved by now

api/pom.xml
KliminV: should be solved

Haarolean:
hey @KliminV,
appreciate the effort on this! while this was open, we ended up migrating to gradle (#783), so this config is no longer needed.

in gradle, these things are now handled like this:
```
release = resolveBooleanProperty("release")
    includeFrontend = resolveBooleanProperty("include-frontend", release)
    buildDockerImages = resolveBooleanProperty("build-docker-images", release)
    runE2e = resolveBooleanProperty("run-e2e")
    e2eHeadlessOff = resolveBooleanProperty("e2e-headless-off")
    e2eSelenoidOff = resolveBooleanProperty("e2e-selenoid-off")
```

if you're up for it, would love to see you tackle another issue - here are a few that might be interesting:
[good first issues](https://github.com/kafbat/kafka-ui/issues?q=is%3Aopen+is%3Aissue+label%3Ascope%2Finfra+label%3A%22good+first+issue%22)

thanks again, and i'll make sure to review quicker next time! :)

PR #671, opened 2024-11-23
feat: Add support for custom root certificates in Java keystore

patsevanton:
### Description
This pull request introduces support for importing custom root certificates into the Java keystore during the Docker image build process. This enhancement allows the application to trust additional certificates, which is particularly useful for environments where custom certificates are required.

### Changes
- **Dockerfile**:
  - Added a step to create a directory for certificates (`/etc/kafkaui/certs`).
  - Added a loop to import any `.crt` files found in the `/etc/kafkaui/certs` directory into the Java keystore using `keytool`.

Please review and provide feedback.

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

### DIFF
--- a/api/import-certs.sh
+++ b/api/import-certs.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+CERT_DIR="/etc/kafkaui/certs"
+KEYSTORE="$JAVA_HOME/lib/security/cacerts"
+STOREPASS="changeit"
+
+if [ -d "$CERT_DIR" ]; then
+    for cert in $CERT_DIR/*.crt; do
+        if [ -f "$cert" ]; then
+            alias=$(basename "$cert" .crt)
+            echo "Importing $cert with alias $alias"
+            keytool -import -noprompt -trustcacerts -alias "$alias" -file "$cert" -keystore "$KEYSTORE" -storepass "$STOREPASS"
+        fi
+    done
+else
+    echo "No certificates directory found at $CERT_DIR"
+fi
+
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -11,15 +11,7 @@ RUN apk add --no-cache \
 RUN addgroup -S kafkaui && adduser -S kafkaui -G kafkaui
 
 # creating folder for dynamic config usage (certificates uploads, etc)
-RUN mkdir -p /etc/kafkaui/certs
-RUN if ls /etc/kafkaui/certs/*.crt 1> /dev/null 2>&1; then \
-        for cert in /etc/kafkaui/certs/*.crt; do \
-            keytool -import -noprompt -trustcacerts -alias $(basename $cert .crt) -file $cert -keystore $JAVA_HOME/lib/security/cacerts -storepass changeit; \
-        done \
-    else \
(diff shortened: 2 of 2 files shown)


patsevanton:
I run `./mvnw clean install -Dmaven.test.skip=true -Pprod`

What commands do I need to run before launching docker build?
Maybe there is a docker-compose in which the docker image is assembled?

api/import-certs.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+CERT_DIR="/etc/kafkaui/certs"
+KEYSTORE="$JAVA_HOME/lib/security/cacerts"
```
Haarolean, line 4: I'd rather see a custom truststore created and passed via these spring properties
```
server:
  ssl:
    trust-store: classpath:keycloak-truststore.jks
    trust-store-password: changeit
```
rather than messing with jre's truststore, what do you think?

Haarolean (changes requested):

Haarolean:
> I run `./mvnw clean install -Dmaven.test.skip=true -Pprod`
> 
> What commands do I need to run before launching docker build? What commands should I use to build a custom docker image? Maybe there is a docker-compose in which the docker image is assembled?

clean install with prod profile enabled will build the image, you don't have to do this manually.

api/import-certs.sh
patsevanton: where configure spring properties?
do I need to add a certificate to keycloak-truststore.jks ?

Haarolean: >do I need to add a certificate to keycloak-truststore.jks ?
we can build a new truststore from scratch within the same script and put it somewhere

>where configure spring properties?
I believe we can try setting env vars like `SERVER_SSL_TRUST-STORE: xxx`, or leave this to the user (given this will be well documented, adding a few config properties is way easier than building a truststore from scratch)

patsevanton: I am not Java developer. I don`t know spring. Could you share simple example without kubernetes?

patsevanton: @Haarolean Could you share simple example without kubernetes?

Haarolean: Sorry for the delay. This is a config property, it has nothing to do with k8s. 
My suggestion is to create a truststore from scratch and have it mounted into the container, so the user can set the location via `SERVER_SSL_TRUST-STORE` themselves. What do you think?

patsevanton: I would like to just specify the k8s secret or k8s configmap with the certificate, and it created the certificate store itself inside

Haarolean:
Given this:
1) you can specify a custom truststore to be used by the app:
```
server:
  ssl:
    trust-store: classpath:keycloak-truststore.jks
    trust-store-password: changeit
```
2) We believe the responsibility of creating a custom truststore is one of the end user and such a custom workflow doesn't quite belong to the repo

We decided to leave it as is.
I've raised an issue (https://github.com/kafbat/ui-docs/issues/33) to document the possibility of adding a custom truststore to the app.
Thanks for your understanding and thanks for your contribution! Feel free to let me know if you have any other questions.

PR #686, opened 2024-11-30
FE: Fix Default Consumer's Lag Value

### LINKED ISSUE #675: FE: Consumers: Lag is displayed as 'N/A' in case of null value
### Issue submitter TODO list

- [X] I've looked up my issue in [FAQ](https://ui.docs.kafbat.io/faq/common-problems)
- [X] I've tried running `main`-labeled docker image and the issue still persists there
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md)

In the "Consumers" tab, there were cases where the LAG value for consumers was N/A.

However, when queried using the `describe` CLI command, LAG was displayed as 0.

I believe that N/A can also be accepted as "Not Available", **so I would like to modify it to display 0.**

### Expected behavior

In the "Consumers" tab, if the consumer's LAG is 0, it is displayed as 0 instead of N/A.

### Your installation details

name: kafka-ui
services:
  kakfa-ui:
    image: provectuslabs/kafka-ui:v0.7.2
    container_name: kakfa-ui
    ports:
      - '80:8080'
    environment:
      - KAFKA_CLUSTERS_0_NAME=kafka_cluster
      - KAFKA_CLUSTERS_0_BOOTSTRAPSERVERS=localhost:29092,localhost:29092,localhost:29092
      - DYNAMIC_CONFIG_ENABLED=true
      - KAFKA_CLUSTERS_0_AUDIT_TOPICAUDITENABLED=true
      - KAFKA_CLUSTERS_0_AUDIT_CONSOLEAUDITENABLED=true

### Steps to reproduce

You can just consume a message to the topic without anything special or start the consumer without any message.

### Screenshots

_No response_

### Logs

_No response_

### Additional context

_No response_

K-Diger:
- [x] **Breaking change?** (if so, please describe the impact and migration path for existing application instances)

close #675 

The default value of Consumer Lag displayed when searching the Consumer List has been modified.

The existing displayed value is 'N/A', which is different from the response value of '0' that the server responds to.

We believe that 'N/A' may mean a state of disuse, so we write it as 0 to make it clear.

What changes did you make? (Give an overview)

Is there anything you'd like reviewers to focus on?

How Has This Been Tested? (put an "x" (case-sensitive!) next to an item)

- [x] No need to
- [ ] Unit checks
- [ ] Integration checks


Haarolean:
`N/A` and `0` values for this field are two distinct things, please see https://github.com/provectus/kafka-ui/issues/3728 for more details. Unless you can provide additional information on why we should treat both the same, I doubt we need these changes. Let me know what you think.

K-Diger:
> `N/A` and `0` values for this field are two distinct things, please see [provectus/kafka-ui#3728](https://github.com/provectus/kafka-ui/issues/3728) for more details. Unless you can provide additional information on why we should treat both the same, I doubt we need these changes. Let me know what you think.이 필드의 N/A 값과 0 값은 서로 다른 두 가지 사항입니다. 자세한 내용은 proctus/kafka-ui#3728을 참조하세요. 두 가지를 동일하게 취급해야 하는 이유에 대한 추가 정보를 제공할 수 없다면 이러한 변경이 필요하지 않을 것입니다. 당신의 생각을 알려주세요.

This is not a situation where the server's response value is lowered to 'N/A'. 

Since the server responds with exactly 0, I think this is different from the situation where it displays 'N/A' for the issue you tagged.

```
{
    "pageCount": 1,
    "consumerGroups": [
        {
            "inherit": "ConsumerGroupDTO",
            "groupId": "connect-secom-alarm-sink-connector",
            "members": 1,
            "topics": 1,
            "simple": false,
            "partitionAssignor": "range",
            "state": "STABLE",
            "coordinator": {
                "id": 1,
                "host": "localhost",
                "port": 29092,
                "bytesInPerSec": null,
                "bytesOutPerSec": null,
                "partitionsLeader": null,
                "partitions": null,
                "inSyncPartitions": null,
                "partitionsSkew": null,
                "leadersSkew": null
            },
            "consumerLag": 0
        }
    ]
}
```

The value displayed in the JSON response above is the same as the photo I attached.

스크린샷 2024-12-06 오전 10 24 23


In conclusion, the tagged issue appears to be a proposal to indicate N/A when data cannot be loaded.

What I am suggesting is to not represent 0 as 'N/A'.

Haarolean:
> Since the server responds with exactly 0, I think this is different from the situation where it displays 'N/A' for the issue you tagged.

In this case, your changes do not make sense to me.
The lag is 0 in your response (`"consumerLag": 0`), which, under current implementation (`return args.getValue() || 'N/A';`) will be displayed as is. `args.getValue()` would be false only in case of `undefined`, not `0`. Am I missing something here?

K-Diger:
> In this case, I don't think your changes make sense. The lag is 0 in your response (`"consumerLag": 0`), which, under current implementation, (`return args.getValue() || 'N/A';`) will be displayed as is. `args.getValue()` would be false only in case of `undefined`, not `0`. Am I missing something here?

I've revised the code again based on your feedback.

This time, the modified code displays the response as 0 only when it is 0.

I modified it to display 'N/A' when the value cannot actually be retrieved, such as null.

K-Diger:
Thanks for your reply. We have made a commit that reflects your feedback,
so please check.

2024년 12월 19일 (목) 오전 12:56, Roman Zabaluev ***@***.***>님이 작성:

K-Diger:
I've added my comments on what you said and a commit for additional test
code.


We would appreciate it if you could give us feedback after checking.

PR #702, opened 2024-12-10
BE: Chore: use enhanced switch

yeikel:
A small change to take advantage of newer language features

- [x] Unit checks

is-this-smile-cute-or-terrifying-photo-u1

### DIFF
--- a/api/src/main/java/io/kafbat/ui/service/KafkaConnectService.java
+++ b/api/src/main/java/io/kafbat/ui/service/KafkaConnectService.java
@@ -207,22 +207,13 @@ public class KafkaConnectService {
   public Mono<Void> updateConnectorState(KafkaCluster cluster, String connectName,
                                          String connectorName, ConnectorActionDTO action) {
     return api(cluster, connectName)
-        .mono(client -> {
-          switch (action) {
-            case RESTART:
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/kafbat/ui/service/KafkaConnectService.java
```diff
@@ -207,22 +207,13 @@ public Mono<Void> deleteConnector(
...
-            case RESTART_FAILED_TASKS:
-              return restartTasks(cluster, connectName, connectorName,
-                  t -> t.getStatus().getState() == ConnectorTaskStatusDTO.FAILED);
-            case PAUSE:
-              return client.pauseConnector(connectorName);
-            case RESUME:
-              return client.resumeConnector(connectorName);
-            default:
-              throw new IllegalStateException("Unexpected value: " + action);
-          }
```
yeikel, line 225: A default case is not needed because we are switching on an enum. If we add new values in the future, the compiler will fail with "'switch' expression does not cover all possible input values"

Haarolean (approved):

yeikel:
@Haarolean Can we merge this? It should be ready to go as far as I know

PR #709, opened 2024-12-13
Infra: FE: Bump pnpm to 9.15.0

yeikel:

Upgrades pnpm to the latest version(9.15.0)

Besides the expected upstream patches in pnpm , it also unblocks dependabot upgrades 

See https://github.com/dependabot/dependabot-core/issues/11124

- [x] Unit checks
- [x] Integration checks

cute-red-panda

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -103,8 +103,8 @@
     "whatwg-fetch": "3.6.20"
   },
   "engines": {
-    "node": "v18.17.1",
-    "pnpm": "v9.11.0"
+    "node": "18.17.1",
+    "pnpm": "9.15.0"
   },
   "pnpm": {
     "overrides": {
--- a/.github/workflows/frontend_tests.yml
+++ b/.github/workflows/frontend_tests.yml
@@ -23,7 +23,7 @@ jobs:
 
       - uses: pnpm/action-setup@v4.0.0
         with:
-          version: 9.11.0
+          version: 9.15.0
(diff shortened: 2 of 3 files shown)


yeikel:
The build failure is due to `pull_request_target` as it is using the version of the workflow from `main` where `9.11.0` is still used

frontend/package.json
```diff
@@ -103,8 +103,8 @@
     "whatwg-fetch": "3.6.20"
   },
   "engines": {
-    "node": "v18.17.1",
-    "pnpm": "v9.11.0"
+    "node": "18.17.1",
+    "pnpm": "9.15.0"
```
yeikel, line 107: My interpretation of the docs https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines is that this should not contain a tag but a version

It seems that some tools read this correctly, but dependabot,for example, doesn't

Haarolean (dismissed):

Haarolean:
@yeikel should we merged given the open discussion in https://github.com/dependabot/dependabot-core/issues/11124 ?

yeikel:
> @yeikel should we merged given the open discussion in https://github.com/dependabot/dependabot-core/issues/11124 ?

Yes

But I think that we should remove the enforcement in a follow-up PR

I can update this one, too, if you prefer

Haarolean (approved):

PR #710, opened 2024-12-13
BE: Chore: Upgrade kafka to confluent 7.8.0

yeikel:
**Description** 

Before this change, we were using both Confluent Artifacts (via `confluent.version`) as well as the OSS artifacts (via `kafka-clients.version` and `kafka.version` ).

With this change, we are upgrading and aligning the versions to use the Confluent Kafka Distribution. This helps us not only to reduce the attack surface but also to simplify the dependency upgrades

As a side effect of this upgrade, we are also extending the retry logic to retry when there is a re-balance in progress as that was previously failing silently as described in https://issues.apache.org/jira/browse/KAFKA-14455

Closes https://github.com/kafbat/kafka-ui/pull/693
Closes https://github.com/kafbat/kafka-ui/pull/692

- [x] Unit checks
- [x] Integration checks

p03t268b

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -50,7 +50,10 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
-            <version>${kafka-clients.version}</version>
+            <!-- ccs stands for Confluent Community Edition
+             See https://www.confluent.io/confluent-community-license-faq/
+             -->
+            <version>${confluent.version}-ccs</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
--- a/api/src/test/java/io/kafbat/ui/AbstractIntegrationTest.java
+++ b/api/src/test/java/io/kafbat/ui/AbstractIntegrationTest.java
@@ -40,7 +40,7 @@ public abstract class AbstractIntegrationTest {
   private static final boolean IS_ARM =
       System.getProperty("os.arch").contains("arm") || System.getProperty("os.arch").contains("aarch64");
 
-  private static final String CONFLUENT_PLATFORM_VERSION = IS_ARM ? "7.2.1.arm64" : "7.2.1";
+  private static final String CONFLUENT_PLATFORM_VERSION = IS_ARM ? "7.8.0.arm64" : "7.8.0";
 
   public static final KafkaContainer kafka = new KafkaContainer(
       DockerImageName.parse("confluentinc/cp-kafka").withTag(CONFLUENT_PLATFORM_VERSION))
--- a/e2e-tests/src/main/java/io/kafbat/ui/screens/panels/NaviSideBar.java
+++ b/e2e-tests/src/main/java/io/kafbat/ui/screens/panels/NaviSideBar.java
@@ -13,6 +13,7 @@ import java.time.Duration;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.openqa.selenium.By;
 
 public class NaviSideBar extends BasePage {
 
@@ -45,7 +46,7 @@ public class NaviSideBar extends BasePage {
   @Step
   public NaviSideBar openSideMenu(String clusterName, MenuItem menuItem) {
     WebUtil.clickByActions(expandCluster(clusterName).parent()
-        .$x(String.format(sideMenuOptionElementLocator, menuItem.getNaviTitle())));
+        .find(By.linkText(menuItem.getNaviTitle())));
     return this;
   }
 
--- a/e2e-tests/src/main/java/io/kafbat/ui/screens/topics/TopicSettingsTab.java
+++ b/e2e-tests/src/main/java/io/kafbat/ui/screens/topics/TopicSettingsTab.java
@@ -9,10 +9,13 @@ import io.kafbat.ui.screens.BasePage;
 import io.qameta.allure.Step;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.NoSuchElementException;
 
 public class TopicSettingsTab extends BasePage {
 
   protected SelenideElement defaultValueColumnHeaderLocator = $x("//div[text() = 'Default Value']");
+  protected SelenideElement nextButton = $x("//button[contains(text(), 'Next')]");
+  protected SelenideElement previousButton = $x("//button[contains(text(), 'Previous')]");
 
   @Step
   public TopicSettingsTab waitUntilScreenReady() {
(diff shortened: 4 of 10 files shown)


api/src/test/java/io/kafbat/ui/AbstractIntegrationTest.java
```diff
@@ -38,7 +38,7 @@ public abstract class AbstractIntegrationTest {
   private static final boolean IS_ARM =
       System.getProperty("os.arch").contains("arm") || System.getProperty("os.arch").contains("aarch64");
 
-  private static final String CONFLUENT_PLATFORM_VERSION = IS_ARM ? "7.2.1.arm64" : "7.2.1";
+  private static final String CONFLUENT_PLATFORM_VERSION = IS_ARM ? "7.8.0.arm64" : "7.8.0";
```
DimaVilda, line 41: hi there
as you noticed you have some failed tests in you MR. They fail because you bumbed confluentinc/cp-kafka version here.
I tested it and here is what I figured out:
1. Test fails randomly. Sometimes it passes, sometimes not. 
2. Such problem starts from '7.5.0' confluentinc/cp-kafka version and further

You can check release notes from: 
https://docs.confluent.io/platform/7.5/release-notes/index.html

I assume the problem related to Kafka Connect initialization. I suggest you to use here 7.4.0 version as last that works and try to ping Confluent guys with this problem.

Haarolean: @DimaVilda @yeikel PTAL at this: https://github.com/kafbat/kafka-ui/pull/573#issuecomment-2464974900, might be related.

DimaVilda: aha, I see
I don't like the workaround, actually and I would vote it will be resolved on Confluent's side, up to you @Haarolean 
In this MR i'd suggest to use stable kafka image version and don't bother with it

yeikel: The comment links to https://issues.apache.org/jira/browse/KAFKA-14455 but that issue was "resolved" and the fix should be included in 7.8

yeikel: Looking at the comment and the issue linked 

> The image upgrade introduced a bug in the tests because some kafka connect errors that were previously processed asynchronously are now rising up to the POST|PUT /connectors responses. See the [KAFKA-14455](https://issues.apache.org/jira/browse/KAFKA-14455) issue.
> 
> Specifically it now randomly produces HTTP 500 errors because Kafka Connect often falls into the "re balancing" state when we create/delete connectors in a row.

I confirmed that this issue started in `7.5.x`. I could not replicate this issue in any version of `7.4.x`.

The reason this issue started is due to the linked JIRA ticket: https://issues.apache.org/jira/browse/KAFKA-14455, which addressed an existing problem in error feedback for Connect. 

As described in the ticket, this was a case of poor user experience where the connector endpoint previously returned `200` even when the call failed asynchronously internally not saving the actual user change and producing a misleading output. With the new change, when a re-balance is in progress, the API returns 500 instead as expected

With this context, I do not believe this is a source of concern. In fact, the reason the tests passed earlier was due to the bad user experience noted in the original issue, where errors were not correctly propagated to users.

I think that moving forward, instead of reporting this upstream, we should update the tests to accommodate the expected behavior, either as part of this pull request or by incorporating the changes implemented by @Dugong42 in his other pull request.

yeikel: Based on the tests I conducted, it seems that most failures occur due to the `@setUp()` method, which attempts to perform another mutation (the connector creation) before the backend has finished processing the previous operation. Essentially, this creates a situation where tests indirectly conflict with each other’s state.

What I am currently trying to determine is what exactly connect is doing that necessitates waiting before proceeding with the next test.

One obvious alternative is to ensure that each test uses its own instance of connect ensuring that we create a clean instance of the dependencies with each test and that no state is shared among tests. The downside of doing that is that the startup of each test will be considerably slower.

yeikel: I updated the code to [retry on Rebalance ](https://github.com/yeikel/kafka-ui/blob/5c1a7e503e8b4733e505dcefcf00fb1a7c96d6ed/api/src/main/java/io/kafbat/ui/client/RetryingKafkaConnectClient.java#L55-L66)and the tests seem more stable right now. That's the solution I believe we should deploy

Haarolean (approved):

api/pom.xml
```diff
@@ -50,7 +50,10 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
-            <version>${kafka-clients.version}</version>
```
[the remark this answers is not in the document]
yeikel: @Haarolean 

The significant version difference is because Confluent follows its own versioning strategy, but ultimately uses the open-source distribution binaries as a foundation. For example, Confluent Platform `v7.8.0` is built on Kafka `v3.8.x`. See this page for more details: https://docs.confluent.io/platform/current/installation/versions-interoperability.html.

In general, it's best to choose one version or the other, as using both together can lead to binary incompatibility issues. For instance, in the current main branch, we're using client versions that pull v3.8.x while also pulling Kafka 3.4.

The only valid reason for using a separate version for the clients package would be if you need to upgrade it independently while delaying the upgrade of other packages. However, this is a limited use case and not really worth the complexity.

For most users, it's better to stick to one version. Mixing them offers few benefits and only increases the workload, as each version will need to be upgraded separately, and you need to be cautious not to mix them up

There are some nices to have that the confluent distribution brings us, but we should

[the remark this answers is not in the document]
Haarolean: Thanks for the explanation. My concern was about introducing incompatible changes with stuff like our serde API, hope that's not the case here.

Haarolean:
@yeikel this e2e test seem to be failing, could you take a look?
`TopicsTest.retentionBytesAccordingToMaxSizeOnDiskCheck:413 » NoSuchElement No value present`

yeikel:
> @yeikel this e2e test seem to be failing, could you take a look? `TopicsTest.retentionBytesAccordingToMaxSizeOnDiskCheck:413 » NoSuchElement No value present`

@Haarolean I updated the test. 

The reason it was failing is because the page was returning 2 pages and the test assumed that all the properties are visible within the first page

e2e-tests/src/main/java/io/kafbat/ui/screens/panels/NaviSideBar.java
```diff
@@ -45,7 +46,7 @@ public String getPagePath(MenuItem menuItem) {
   @Step
   public NaviSideBar openSideMenu(String clusterName, MenuItem menuItem) {
     WebUtil.clickByActions(expandCluster(clusterName).parent()
-        .$x(String.format(sideMenuOptionElementLocator, menuItem.getNaviTitle())));
+        .find(By.linkText(menuItem.getNaviTitle())));
```
yeikel, line 49: This selector seems to be more stable as the previous selector was failing in my environment

e2e-tests/src/main/java/io/kafbat/ui/screens/topics/TopicSettingsTab.java
```diff
@@ -36,7 +39,25 @@ private TopicSettingsTab.SettingsGridItem getItemByKey(String key) {
 
   @Step
   public String getValueByKey(String key) {
-    return getItemByKey(key).getValue();
+    while (true) {
```
yeikel, line 42: When the key is not in the first page, we need to navigate page by page to find the value

Haarolean (approved):

Haarolean:
@yeikel thank you very much for this PR!

Issue #707, opened 2024-12-12
Infra: Consider running the `Infra: CVE checks` with every commit as well


yeikel:
### Issue submitter TODO list

- [X] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md) and the feature is not present there

### Is your proposal related to a problem?

Currently, the "Infra: CVE checks" check is configured to run [twice per month](https://github.com/kafbat/kafka-ui/actions/workflows/cve_checks.yml), and while that is great, it does not raise the constant awareness that CVE should raise. 

For example, the [latest run failed ](https://github.com/kafbat/kafka-ui/actions/runs/12249665806/job/34171351979) but it is only known if we navigate to the specific build check while `main` is considered to be healthy

### Describe the feature you're interested in

We should extend the `cve_checks.yml` workflow to also run on pull requests and merges to `main`. It should continue to be a separate check

- Pull requests: Because we should not be introducing new dependencies with CVEs 
- On Main : To raise awareness and serve as a continuous reminder that actions may be needed

### Describe alternatives you've considered

Use the existing schedule and remember to check manually

### Version you're running

318bcc91ddcca72b6a280d20f5f51ea85dac77c0

Haarolean:
>Pull requests: Because we should not be introducing new dependencies with CVEs

I don't mind this one, would be nice to know if there's anything new in case of newly introduced dependencies.

>On Main : To raise awareness and serve as a continuous reminder that actions may be needed

This one is questionable, considering we don't have much bandwidth to fix these issues too often. Like, #256 was open for half a year. Running this on every commit sounds like an extra noise.

@yeikel what do you think?

yeikel:
> > Pull requests: Because we should not be introducing new dependencies with CVEs
> 
> I don't mind this one, would be nice to know if there's anything new in case of newly introduced dependencies.
> 
> > On Main : To raise awareness and serve as a continuous reminder that actions may be needed
> 
> This one is questionable, considering we don't have much bandwidth to fix these issues too often. Like, #256 was open for half a year. Running this on every commit sounds like an extra noise.
> 
> @yeikel what do you think?

I agree that it would be noise if it not actioned, but I think that this is _good_ noise. We should be more than aware than issues with CVE exist and I think that every commit is a good way to show that

Haarolean:
[...]

PR #745, opened 2025-01-03
Infra: Run cve checks workflow on pull requests and pushes to main

yeikel:

Closes https://github.com/kafbat/kafka-ui/issues/707

Tested here : 

- `main` : https://github.com/yeikel/kafka-ui/actions/runs/12602983150/job/35127088622
- Pull request: https://github.com/yeikel/kafka-ui/actions/runs/12603014749/job/35127194824?pr=68

346617360_1196147131002766_412235603490618042_n

### DIFF
--- a/.github/workflows/cve_checks.yml
+++ b/.github/workflows/cve_checks.yml
@@ -1,5 +1,8 @@
 name: "Infra: CVE checks"
 on:
+  pull_request:
+  push:
+    branches: [ "main" ]
   workflow_dispatch:
   schedule:
     # * is a special character in YAML so you have to quote this string
@@ -71,7 +74,7 @@ jobs:
 
   notify:
     needs: check-cves
-    if: ${{ always() && needs.build-and-test.result == 'failure' }}
+    if: ${{ always() && needs.build-and-test.result == 'failure' && github.event_name == 'schedule' }}
     uses: ./.github/workflows/infra_discord_hook.yml
     with:
       message: "Attention! CVE checks run failed! Please fix them CVEs :("
--- a/.github/workflows/cve_checks.yml
+++ b/.github/workflows/cve_checks.yml
@@ -1,6 +1,7 @@
 name: "Infra: CVE checks"
 on:
   pull_request:
+    types: [ "opened", "reopened", "synchronize" ]
   push:
     branches: [ "main" ]
   workflow_dispatch:


yeikel:
@Haarolean I noticed that you implemented this recently: https://github.com/kafbat/kafka-ui/pull/219

We probably do not want to send notifications for every failure (definitely not from pull requests). Do you have any suggestion about when should it be sent?

.github/workflows/cve_checks.yml
```diff
@@ -1,5 +1,8 @@
 name: "Infra: CVE checks"
 on:
+  pull_request_target:
```
Haarolean, line 3: I think we don't need pull_request_target for this particular workflow, `pull_request` should suffice.

.github/workflows/cve_checks.yml
```diff
@@ -71,7 +74,7 @@ jobs:
 
   notify:
     needs: build-and-test
-    if: ${{ always() && needs.build-and-test.result == 'failure' }}
+    if: ${{ always() && needs.build-and-test.result == 'failure' && github.ref == 'refs/heads/main' }}
```
Haarolean, line 77: Could you please adjust this so it only runs upon failures of scheduled runs?

Haarolean (changes requested):

Haarolean:
> Do you have any suggestion about when should it be sent?

Thank you for this PR.
I'd suggest to send these only in case of a failed *scheduled* run, would you be able to adjust the workflow this way?

.github/workflows/cve_checks.yml
Haarolean: @yeikel 👀

yeikel: No, I think that we should use `pull_request_target`. If we add `pull_request` a malicious actor can still modify this workflow and potentially leak secrets from the project. In fact, we should probably have `pull_request_target` for all the workflows

yeikel:
> > Do you have any suggestion about when should it be sent?
> 
> Thank you for this PR. I'd suggest to send these only in case of a failed _scheduled_ run, would you be able to adjust the workflow this way?

I addressed this with https://github.com/kafbat/kafka-ui/commit/94da91b522a0af1dc13f6cc7d1f5232cab143fc2

it should be ready to merge now I believe

Thanks!

.github/workflows/cve_checks.yml
Haarolean: It's the other way around - the target one grants an access to secrets

yeikel: Looking a bit more into it, it seems that you are right and that `pull_request` is generally safer

yeikel: Namely this part 

> only use pull_request_target for workflows where access to secrets or write permissions is strictly necessary, such as commenting, labeling, or status updates on pull requests.

yeikel: I updated it with https://github.com/kafbat/kafka-ui/pull/745/commits/1300d0c52800d3ff1936fee05bf21759181be1ea

Thanks!

.github/workflows/cve_checks.yml
```diff
@@ -1,5 +1,8 @@
 name: "Infra: CVE checks"
 on:
+  pull_request:
```
Haarolean, line 3: This one implies specifying types, for example we don't need to re-run this on `labeled` event

Haarolean (changes requested):

.github/workflows/cve_checks.yml
yeikel: Updated it with https://github.com/kafbat/kafka-ui/pull/745/commits/fcd6acbd43b38e6978f288537d43bb61ba58c8d2

PR #748, opened 2025-01-03
Infra: FE: migrate to node 22

yeikel:

Node 18 reached end of life in 18 Oct 2023. The latest LTS version of Node is 22.x

See https://endoflife.date/nodejs

- [x] Unit checks
- [x] Integration checks

cute-wild-animals-photos-hardcore-aww-2-63345025740e4__700


frontend/package.json
```diff
@@ -103,7 +103,7 @@
     "whatwg-fetch": "3.6.20"
   },
   "engines": {
-    "node": "18.17.1",
+    "node": ">=22.12.0",
```
Haarolean, line 106: Sorry I'm not quite familiar with how this works, won't we get any potential issues if versions like 23.x will be used one day?

yeikel: You are right in your suspicions, thank you! 

I updated it https://github.com/kafbat/kafka-ui/pull/748/commits/ba033d63cd6eb228a0965f20fe15278017b44d0a to use the `^`  to only allow the current major version: 

    Greater than or equal to 22.12.0
    Less than 23.0.0 (the next major version)

Please note that in our current configuration, we are not strictly enforcing this so it only produces a log message: 


>  WARN  Unsupported engine: wanted: {"node":"^22.12.0"} (current: {"node":"v23.5.0","pnpm":"9.15.3"})
> Lockfile is up to date, resolution step is skipped

If the intention was to block installation, we would need to use Strict engine checks.

Haarolean: >If the intention was to block installation, we would need to use Strict engine checks.
I believe the only point for limiting the values here is not to get things broken by an unintentional node upgrade, not sure limiting major versions makes sense, perhaps release ones is enough. What do you think?

yeikel: I think that what we have right now is better (warn, but not fail, suggest current major version). Generally, it is safe to test within the current major version

If we strictly enforce a version, it will break tools like dependabot(at least for now)

yeikel:
@Haarolean Can you approve the other workflows?

Haarolean:
> @Haarolean Can you approve the other workflows?

not sure that's needed since we're adjusting a workflow here and it won't run until merged, so we can't test this thing before merging anyway, unless run on a fork.

Haarolean (approved):

Haarolean:
@yeikel thank you!

PR #704, opened 2024-12-12
refactor: remove unnecessary architecture check in unit test

yeikel:
**What changes did you make?** (Give an overview)

The existing logic to check for arm vs non-arm is unnecessary because the `confluentinc/cp-kafka` image is a [multi-platform image](https://docs.docker.com/build/building/multi-platform/) and the docker engine will pick the appropriate version as needed

**How Has This Been Tested?** (put an "x" (case-sensitive!) next to an item)

- [x] Unit checks
- [x] Integration checks
- [x] Covered by existing automation

**Checklist** (put an "x" (case-sensitive!) next to all the items, otherwise the build will fail)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (e.g. **ENVIRONMENT VARIABLES**)
- [ ] My changes generate no new warnings (e.g. Sonar is happy)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged

**A picture of a cute animal (not mandatory but encouraged)**

PR #771, opened 2025-01-14
Docs: Remove references to arm-specific local setup

yeikel:

This should not be needed. See https://github.com/kafbat/kafka-ui/pull/704

This pull request is accompanied by the following doc update https://github.com/kafbat/ui-docs/pull/35

22202386d4415bf4ea3648022237ccd1

### DIFF
--- a/.dev/dev.yaml
+++ b/.dev/dev.yaml
@@ -1,9 +1,3 @@
-# This is a compose file designed for arm64/Apple Silicon systems
-# To adapt this to x86 please find and replace ".arm64" with empty
-
-# ARM64 supported images for kafka can be found here
-# https://hub.docker.com/r/confluentinc/cp-kafka/tags?page=1&name=arm64
----
 version: '3.8'
 name: "kafbat-ui-dev"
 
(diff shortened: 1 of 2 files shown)


.dev/dev_arm64.yaml
```diff
@@ -1,132 +0,0 @@
-# This is a compose file designed for arm64/Apple Silicon systems
-# To adapt this to x86 please find and replace ".arm64" with empty
-
-# ARM64 supported images for kafka can be found here
-# https://hub.docker.com/r/confluentinc/cp-kafka/tags?page=1&name=arm64
----
-version: '3.8'
-name: "kafbat-ui-dev"
```
Haarolean, line 8: Let's keep this compose please. It's a generic compose for dev purposes without any specific stuff other compose files provide (sasl, proxy, etc). We can remove the 'arm64' postfix in image names and rename the file instead.

Haarolean (changes requested):

.dev/dev_arm64.yaml
yeikel: I updated it.

Haarolean (approved):

Haarolean:
Thank you

PR #773, opened 2025-01-14
BE: upgrade logback to 1.5.16

yeikel:
**What changes did you make?** (Give an overview)

Fixes CVE-2024-12798 and CVE-2024-12801 while we wait for the next Spring Boot Release

**How Has This Been Tested?** (put an "x" (case-sensitive!) next to an item)

- [x] Covered by existing automation
- [x] Manual tests 

Cherry-picked 150b7af31a70c25c46947102df04acfd09a0f988 to test with changes in https://github.com/kafbat/kafka-ui/pull/745 

Without this PR : https://github.com/kafbat/kafka-ui/actions/runs/12760141709/job/35565086663?pr=745

image

With this PR:  https://github.com/kafbat/kafka-ui/actions/runs/12760148833/job/35565105573?pr=773

image

**Checklist** (put an "x" (case-sensitive!) next to all the items, otherwise the build will fail)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

**A picture of a cute animal (not mandatory but encouraged)**

baby-pangolin-on-mother

PR #783, opened 2025-01-15
Infra: Add gradle p.1

germanosin:

- [ ] No need to
- [] Unit checks
- [ ] Integration checks


api/build.gradle
```diff
@@ -0,0 +1,115 @@
...
+plugins {
+    id 'antlr'
+    id 'checkstyle'
+    id 'org.springframework.boot' version '3.4.1'
+    id 'com.gorylenko.gradle-git-properties' version "2.4.2"
+}
+
+dependencies {
+    implementation project(":contract")
+    implementation project(":frontend")
```
Haarolean, line 11: api shouldn't depend on the frontend

api/build.gradle
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+tasks.withType(Checkstyle) {
+    exclude '**/ksql/**'
+}
+
+checkstyle {
+    toolVersion '10.3.1'
+    configFile = rootProject.file('etc/checkstyle/checkstyle-e2e.xml')
```
Haarolean, line 107:
```suggestion
    configFile = rootProject.file('etc/checkstyle/checkstyle.xml')
```
due to e2e having a different rule set

contract/build.gradle
```diff
@@ -0,0 +1,116 @@
+import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
+
+plugins {
+    id "java-library"
+    id 'org.openapi.generator' version '7.9.0'
```
Haarolean, line 5: can we extract this version as well to a version catalog (gradle's "dependency management" list) to make it possible for dependabot to track these versions as well?

e2e-tests/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+
+    implementation libs.lombok
+
+    annotationProcessor libs.lombok
+    testAnnotationProcessor libs.lombok
+}
+
+checkstyle {
+    toolVersion '10.3.1'
+    configFile = rootProject.file('etc/checkstyle/checkstyle.xml')
```
Haarolean, line 34:
```suggestion
    configFile = rootProject.file('etc/checkstyle/checkstyle-e2e.xml')
```

gradle.properties
```diff
@@ -0,0 +1,7 @@
+description='Kafka UI'
```
Haarolean, line 1:
```suggestion
description='Kafbat UI'
```

gradle.properties
```diff
@@ -0,0 +1,7 @@
...
+version=0.1.0-SNAPSHOT
```
Haarolean, line 2:
```suggestion
version=0.0.1-SNAPSHOT
```
to match the one in maven's pom

gradle.properties
```diff
@@ -0,0 +1,7 @@
...
+group=io.kafbat.ui
+local_node = false
+node_version = 18.17.1
```
Haarolean, line 5:
```suggestion
node_version = 22.12.0
```
as per aaac4d7c87765303b6b39c2a2a43ee307ce7ab2a

gradle/libs.versions.toml
```diff
@@ -0,0 +1,103 @@
+[versions]
```
Haarolean, line 1: @yeikel I believe this answers your question regarding the version catalog

api/build.gradle
```diff
@@ -0,0 +1,115 @@
+
+plugins {
+    id 'antlr'
+    id 'checkstyle'
+    id 'org.springframework.boot' version '3.4.1'
```
Haarolean, line 5: `gradle/libs.versions.toml` has 3.3.6

gradle/libs.versions.toml
```diff
@@ -0,0 +1,103 @@
...
+spring-boot = '3.3.6'
+spring-ldap = '6.3.5'
```
Haarolean, line 3: why `6.3.2`? in maven we don't override this and the current version is `6.4.2`

serde-api/build.gradle
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+    }
+}
+
+//nexusPublishing {
+//    repositories {
+//        sonatype {
+//            nexusUrl = uri("https://s01.oss.sonatype.org/service/local/")
+//            snapshotRepositoryUrl = uri("https://s01.oss.sonatype.org/content/repositories/snapshots/")
```
Haarolean, line 96: ? 👀

serde-api/gradle.properties
```diff
@@ -0,0 +1,2 @@
+sonatypeUsername=test
```
Haarolean, line 1: do we need to commit these?

Haarolean (changes requested):

api/build.gradle
germanosin: On prod build should

gradle.properties
germanosin: 0.1.0 is more correct version in this case

gradle.properties
germanosin: pom use 18 version

serde-api/gradle.properties
germanosin: Yes, we need any value to be present during build time

api/build.gradle
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+}
+
+tasks.withType(Checkstyle) {
+    exclude '**/ksql/**'
+}
+
+checkstyle {
+    toolVersion '10.3.1'
```
yeikel, line 108: Can this be added to the catalog?

e2e-tests/build.gradle
```diff
@@ -0,0 +1,48 @@
+plugins {
+    id 'java'
+    id 'checkstyle'
+    alias(libs.plugins.allure)
+}
+
+ext {
+    kafkaVersion = "3.8.0"
+    aspectjVersion = "1.9.21"
+    allureVersion = "2.27.0"
```
yeikel, line 10: Same for these

yeikel: Also, this version does not seem correct. It should be from the confluent distribution

e2e-tests/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+}
+
+dependencies {
+    implementation project(":contract")
+    implementation "org.apache.kafka:kafka_2.13:${kafkaVersion}"
+    implementation "org.aspectj:aspectjweaver:${aspectjVersion}"
+    implementation "commons-io:commons-io:2.16.1"
+    implementation "org.testng:testng:7.10.0"
+    implementation "com.codeborne:selenide:7.2.3"
```
yeikel, line 19: Let's please add this to the catalog as well

serde-api/gradle.properties
yeikel: Can any value be ""? `test` is a bit confusing

api/build.gradle
```diff
@@ -0,0 +1,130 @@
...
+    implementation libs.apache.datasketches
+
+    implementation libs.confluent.schema.registry.client
+    implementation libs.confluent.avro.serializer
+    implementation libs.confluent.protobuf.serializer
+    implementation libs.confluent.json.schema.serializer
+
+    implementation libs.aws.msk.auth
+    implementation (libs.azure.identity) {
+        exclude group: 'io.netty', module: 'netty-tcnative-boringssl-static'
```
yeikel, line 40: Let's please explain why we are excluding this

gradle/libs.versions.toml
yeikel: It does, but we need to use it more extensively based on my feedback above

yeikel:
Please update the dependabot configuration as well 🙏  

See https://github.com/kafbat/kafka-ui/blob/ed49499241b429da5c8372e8cd4b100f22005e88/.github/dependabot.yml#L4

serde-api/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+
+            from components.java
+
+            artifact(sourceJar) {
+                classifier = 'sources'
+            }
+
+            artifact(javadocJar) {
+                classifier = 'javadoc'
+            }
```
yeikel, line 54: Is this needed? Do we publish our JAR to central, or who consumes us as JAR directly?

api/build.gradle
```diff
@@ -0,0 +1,130 @@
...
+}
+
+test {
+    useJUnitPlatform()
+}
+
+jib {
+    from {
+        image = 'azul/zulu-openjdk-alpine:21.0.5-jre-headless'
+    }
```
yeikel, line 122: Can this be delegated to a Dockerfile? 

The main downside of setting this here is that dependabot won't get any visibility and we will need to update it manually

serde-api/gradle.properties
Haarolean: I agree that it's a bit confusing

api/build.gradle
Haarolean: hmmm not quite, in maven there's a `skipUIBuild` flag to skip UI build even in prod profile. Let's address this functionality

gradle.properties
Haarolean: why? in our poms it's 0.0.1

serde-api/build.gradle
Haarolean: yes we do
https://central.sonatype.com/artifact/io.kafbat.ui/serde-api

api/build.gradle
```diff
@@ -0,0 +1,130 @@
...
+    maxErrors = 0
+}
+
+test {
+    useJUnitPlatform()
+}
+
+jib {
+    from {
+        image = 'azul/zulu-openjdk-alpine:21.0.5-jre-headless'
```
Haarolean, line 121: I see a few major issues with using jib:
1. Currently, in our [Dockerfile](https://github.com/kafbat/kafka-ui/blob/main/api/Dockerfile#L6) we have a few additional packages installed:
- `gcompat` required for snappy 
- `tzdata` as requested by someone in the past to be able to configure timezones properly in the container
2. We're creating a `/etc/kafkaui/` dir for a dynamic config. I am unsure what will happen with the current implementation, dynamic config operations might fail.
3. We're running the app as a non-root user there, not sure what's the default behavior with jib?
4. Currently dependabot can properly bump the base image version present in dockerfile, however, it won't be able to work with this version located here, in `api/build.gradle`.

Overall, I prefer dockerfile build with a mvn/gradle plugin rather than jib in favor of more granular customization possibilities.

Haarolean (changes requested):
Also, in maven we have `prod` and `local` profiles, the latter aimed at skipping frontend build. Can we address this as well please?

api/build.gradle
yeikel: Plus the lack of dependabot support mentioned above

serde-api/build.gradle
Haarolean: ikr? one of the two hardest things in computer science

contract/build.gradle
germanosin: Ready

e2e-tests/build.gradle
germanosin: This is e2e we'll make it more "gradle-style" later, current task is to migrate

e2e-tests/build.gradle
germanosin: This is e2e we'll make it more "gradle-style" later, current task is to migrate

api/build.gradle
germanosin: Same as in maven

api/build.gradle
germanosin: Let's change dependabot after full migration as a separate task

e2e-tests/build.gradle
yeikel: I am not following. Is the current e2e implementation using their own versions?

Migration shouldn't degrade the quality compared to main

I don't believe we are in a rush to migrate

api/build.gradle
yeikel: Change in what way? Dependabot doesn't support this

serde-api/build.gradle
yeikel: So, we have consumers that use the API code somehow but not the UI? Any idea about what the use case is?

Haarolean: @yeikel there are two things:
1) serde-api is the package to implement custom serdes, like https://github.com/kafbat/ui-serde-glue/
2) some people might use the app without a need for the frontend, to automate some stuff, like tests, that's a valid use case

api/build.gradle
Haarolean: not sure why'd we need skipUIBuild having a local profile, resolving this

yeikel:
What's the status of this PR? I'd love an update given that it blocks https://github.com/kafbat/kafka-ui/pull/773

Haarolean:
> What's the status of this PR? I'd love an update given that it blocks #773
> 
> Can we merge #773 until this is completed? The conflict should be very minimal and easy to resolve anyways

This will take some time. Let's discuss this in #773

Haarolean (approved):

api/pom.xml
```diff
@@ -456,10 +456,10 @@
                                     <goal>copy-resources</goal>
                                 </goals>
                                 <configuration>
-                                    <outputDirectory>${basedir}/target/classes/static</outputDirectory>
+                                    <outputDirectory>${basedir}/target/classes/</outputDirectory>
                                     <resources>
                                         <resource>
-                                            <directory>../frontend/build</directory>
+                                            <directory>../frontend/build/vite/static</directory>
```
wernerdv, line 462: After this change, if you build the project as before using Maven and the command `mvn clean install -Pprod -DskipTests`, then attempting to open the page in the browser will result in an error:

```
2025-03-09 06:35:37,493 ERROR [parallel-4] o.s.b.a.w.r.e.AbstractErrorWebExceptionHandler: [4a5b796b-1]  500 Server Error for HTTP GET "/"  
2025-03-09T06:35:37.499916272Z java.io.FileNotFoundException: class path resource [static/index.html] cannot be opened because it does not exist  
2025-03-09T06:35:37.499924293Z 	at org.springframework.core.io.ClassPathResource.getInputStream(ClassPathResource.java:215) 
```

@Haarolean Could you please let me know if there are plans to support Maven builds during some transition period, or will the files related to Maven be removed in the near future?

Haarolean: @wernerdv gradle migration is complete though, you can achieve the same this way:
```
./gradlew clean build -x test -Pinclude-frontend=true -Pbuild-docker-images=true
```
The rest of maven-related files are to be removed. Let me know if you experience any issues with building via gradle.

PR #726, opened 2024-12-26
BE: RBAC: Add integration test for Active Directory auth

wernerdv:
- [ ] **Breaking change?** (if so, please describe the impact and migration path for existing application instances)

**What changes did you make?** (Give an overview)
Added integration test for Active Directory auth.
At the moment, it only works if `rbac.roles.name` is group from AD.

**Is there anything you'd like reviewers to focus on?**

**How Has This Been Tested?** (put an "x" (case-sensitive!) next to an item)

- [ ] No need to
- [ ] Manually (please, describe, if necessary)
- [ ] Unit checks
- [x] Integration checks
- [ ] Covered by existing automation

**Checklist** (put an "x" (case-sensitive!) next to all the items, otherwise the build will fail)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (e.g. **ENVIRONMENT VARIABLES**)
- [x] My changes generate no new warnings (e.g. Sonar is happy)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged

Check out [Contributing](https://github.com/kafbat/kafka-ui/blob/main/.github/CONTRIBUTING.md) and [Code of Conduct](https://github.com/kafbat/kafka-ui/blob/main/.github/CODE-OF-CONDUCT.md)

**A picture of a cute animal (not mandatory but encouraged)**

PR #779, opened 2025-01-15
LDAP: Fix property nullability condition

Haarolean:
- [ ] **Breaking change?** (if so, please describe the impact and migration path for existing application instances)

**What changes did you make?** (Give an overview)
Fixes #778

**Is there anything you'd like reviewers to focus on?**

**How Has This Been Tested?** (put an "x" (case-sensitive!) next to an item)

- [ ] No need to
- [ ] Manually (please, describe, if necessary)
- [ ] Unit checks
- [ ] Integration checks
- [ ] Covered by existing automation

**Checklist** (put an "x" (case-sensitive!) next to all the items, otherwise the build will fail)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (e.g. **ENVIRONMENT VARIABLES**)
- [ ] My changes generate no new warnings (e.g. Sonar is happy)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged

Check out [Contributing](https://github.com/kafbat/kafka-ui/blob/main/.github/CONTRIBUTING.md) and [Code of Conduct](https://github.com/kafbat/kafka-ui/blob/main/.github/CODE-OF-CONDUCT.md)

**A picture of a cute animal (not mandatory but encouraged)**

PR #812, opened 2025-02-02
feat(LDAP): add integration tests for LDAP Authorization

### LINKED ISSUE #782: BE: Impl LDAP integration test
To prevent issues like #779, we'd need a basic LDAP integration test to test the authentication. Something like #726

sixdouglas:

This is a PR to add Integration tests for the LDAP Authorization capability. Based on the work done for Active Directory in #726 

closes #782

- [ ] No need to
- [ ] Unit checks
- [X] Integration checks


api/src/test/resources/application-rbac-ldap.yml
```diff
@@ -0,0 +1,38 @@
...
+  ldap:
+    base: "cn={0},ou=people,dc=kafbat,dc=io"
+    admin-user: "cn=admin,dc=kafbat,dc=io"
+    admin-password: "StrongPassword123"
+    user-filter-search-base: "dc=kafbat,dc=io"
+    user-filter-search-filter: "(&(uid={0})(objectClass=inetOrgPerson))"
+    group-filter-search-base: "ou=people,dc=kafbat,dc=io" # required for RBAC
+oauth2:
+  ldap:
+    activeDirectory: false
```
Haarolean, line 11: this can be removed as it should default to false anyway, might be a good thing to catch possible issues like in https://github.com/kafbat/kafka-ui/pull/779/files

api/src/test/java/io/kafbat/ui/OpenLdapPIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Test
+  public void testDirectUserPermissions() {
+    AuthenticationInfoDTO info = authenticationInfo("jacksmith");
+
+    assertNotNull(info);
+    assertTrue(info.getRbacEnabled());
+    System.out.println("info = " + info);
+    List<UserPermissionDTO> permissions = info.getUserInfo().getPermissions();
+    assertFalse(permissions.isEmpty());
```
Haarolean, line 80: this particular test looks redundant -- in `testUserPermissions` we've asserted that permissions of `jacksmith` are the same as `johndoe`'s **and** have a topic permissions

api/src/test/java/io/kafbat/ui/OpenLdapPIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+  public void testUserPermissions() {
+    AuthenticationInfoDTO info = authenticationInfo("johndoe");
+
+    assertNotNull(info);
+    assertTrue(info.getRbacEnabled());
+    System.out.println("info = " + info);
+    List<UserPermissionDTO> permissions = info.getUserInfo().getPermissions();
+    assertFalse(permissions.isEmpty());
+    assertTrue(permissions.stream().anyMatch(permission ->
+        permission.getClusters().contains(LOCAL) && permission.getResource() == ResourceTypeDTO.TOPIC));
```
Haarolean, line 67: can we also check that the `action` of the permission is one of `ALL` enum? (should be true IIRC)

api/src/test/java/io/kafbat/ui/OpenLdapPIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+        .getResponseBody()
+        .blockFirst();
+  }
+
+  public static class Initializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+
+    @Override
+    public void initialize(ConfigurableApplicationContext context) {
+      System.setProperty("spring.ldap.urls", LDAP_CONTAINER.getLdapUrl());
+      System.setProperty("oauth2.ldap.activeDirectory", "false");
```
Haarolean, line 126: perhaps redundant as mentioned in other comment

Haarolean (changes requested):
@sixdouglas thank you very much for raising this PR, LDAP tests are much needed indeed :)

api/src/test/resources/application-rbac-ldap.yml
sixdouglas: Ok, you're right, let's remove redundant config settings

api/src/test/java/io/kafbat/ui/OpenLdapPIntegrationTest.java
sixdouglas: Remove too, thanks for the catch

api/src/test/java/io/kafbat/ui/OpenLdapPIntegrationTest.java
sixdouglas: I've added a test for this, would you like also to add a more fine grained test with a really specific right on just one topic?

Haarolean: perhaps there's no need as we should (and have now) tests specifically for RBAC itself. Thanks!

sixdouglas:
@Haarolean removing both config: 
```
System.setProperty("oauth2.ldap.activeDirectory", "false");
```
and this one: 
```
oauth2:
  ldap:
    activeDirectory: false
```
Break the LDAP tests in the CI, though it works on my machine (I never though I would use this sentence again ;) ).
I'll keep looking for a reason and solution

Haarolean:
> @Haarolean removing both config: 
> 
> ```
> 
> System.setProperty("oauth2.ldap.activeDirectory", "false");
> 
> ```
> 
> and this one: 
> 
> ```
> 
> oauth2:
> 
>   ldap:
> 
>     activeDirectory: false
> 
> ```
> 
> Break the LDAP tests in the CI, though it works on my machine (I never though I would use this sentence again ;) ).
> 
> I'll keep looking for a reason and solution

I'll take a look as well, might be faster

Haarolean:
@sixdouglas It also fails locally when all tests are run (`mvn clean test`) rather than a single test. It's an issue with the context

sixdouglas:
@Haarolean It looks like there is some sort of Spring context conflict between LdapTest and ActiveDirectoryTest. 
In order to make tests pass every-time, I declared the context as dirty for each of these three tests since test execution order is random.

I don't know why the build failed for this test: 
```
SchemaRegistryServiceTests.shouldReturnNotEmptyResponseWhenGetAllSchemas:272->lambda$shouldReturnNotEmptyResponseWhenGetAllSchemas$3:281 » NoSuchElement No value present
```

Haarolean:
> @Haarolean It looks like there is some sort of Spring context conflict between LdapTest and ActiveDirectoryTest. In order to make tests pass every-time, I declared the context as dirty for each of these three tests since test execution order is random.
> 
> I don't know why the build failed for this test:
> 
> ```
> SchemaRegistryServiceTests.shouldReturnNotEmptyResponseWhenGetAllSchemas:272->lambda$shouldReturnNotEmptyResponseWhenGetAllSchemas$3:281 » NoSuchElement No value present
> ```

not a fan of `@DirtiesContext`, slapping this on every test that's interfering with some other one gonna cost us a lot in the future: emerging number of tests will increase the overall test run time (reloading a context takes time)

Issue #850, opened 2025-02-19
BE: RBAC: Support provider for basic auth


Haarolean:
Implement RBAC support for basic auth provider

Pros:
- more control than a r/o filter
- no need to integrate oauth provider like keycloak for a single user installation
- no need to support r/o filter anymore?:)

wernerdv:
Please assign this task to me.

Haarolean:
@wernerdv kind reminder

Haarolean:
Seems redundant once #344 is done. Please follow and upvote #344 if you're interested.

Issue #344, opened 2024-05-03
RBAC: Implement instance-wide default role


seono:
### Issue submitter TODO list

- [X] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md) and the feature is not present there

### Is your proposal related to a problem?

Hi,
When new kafka consumer gives me a request to approach kafka ui (for readonly),
i have to rewrite rbac yaml file.
I deployed two Kafka UI to solve this problem.
First one has basic authentication is for admin users.
The other one is readonly for kafka consumers.

### Describe the feature you're interested in

I think it is useful if the way i can setup default role is exists.
```yaml
rbac:
  roles:
    - name: "admins"
      clusters:
        - development
      subjects:
         - provider: oauth_github
           type: user
           value: 'seono'
      permissions:
        ...
    - name: "readonly"
      clusters:
        - development
      subjects:
        - provider: oauth_github
          type: default
      permissions:
        ...
```

### Describe alternatives you've considered

I know there is other type `organization` and `team`.
If someone wants readonly permission, I can invite them to the group.
But.. In my case, my company has too many organzations and teams so i can not use it properly.

### Version you're running

https://github.com/kafbat/kafka-ui/commit/9843ee0

sankhyan:
hi @Haarolean any update on this feature, it's very important for us since default role is needed for company wide

Haarolean:
@sankhyan As development is done in the spare time of our team members, our bandwidth is limited. You (or your company) can sponsor the development of this feature: https://github.com/sponsors/kafbat/sponsorships?tier_id=384551

githubwalvarado:
Being able to use regex will also help in my particular case.

Haarolean:
@githubwalvarado that's a diff issue: #300

Haarolean:
TODO:
- Implement a default role with this config syntax:
```
rbac:
  default-role: "memelords"
  roles: []
```
- The role should be assigned to anyone who can successfully login

wernerdv:
I'll prepare the PR.

seono:
@Haarolean @wernerdv 
hi, I wanted to give it a try, so I created a draft PR https://github.com/kafbat/kafka-ui/pull/1056


> TODO:
> 
> * Implement a default role with this config syntax:
> 
> ```
> rbac:
[...]

PR #917, opened 2025-03-11
BE: RBAC: Support provider for basic auth

wernerdv:

Resolve https://github.com/kafbat/kafka-ui/issues/850

- [ ] No need to
- [ ] Unit checks
- [x] Integration checks

### DIFF
--- a/api/src/main/java/io/kafbat/ui/config/auth/BasicAuthSecurityConfig.java
+++ b/api/src/main/java/io/kafbat/ui/config/auth/BasicAuthSecurityConfig.java
@@ -1,22 +1,40 @@
 package io.kafbat.ui.config.auth;
 
+import io.kafbat.ui.model.rbac.Role;
+import io.kafbat.ui.model.rbac.provider.Provider;
+import io.kafbat.ui.service.rbac.AccessControlService;
 import io.kafbat.ui.util.StaticFileWebFilter;
+import java.util.Collection;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.security.SecurityProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.HttpMethod;
 import org.springframework.security.config.annotation.web.reactive.EnableWebFluxSecurity;
 import org.springframework.security.config.web.server.SecurityWebFiltersOrder;
 import org.springframework.security.config.web.server.ServerHttpSecurity;
+import org.springframework.security.core.userdetails.ReactiveUserDetailsService;
+import org.springframework.security.core.userdetails.User;
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.web.server.SecurityWebFilterChain;
 import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatchers;
+import org.springframework.util.StringUtils;
+import reactor.core.publisher.Mono;
 
 @Configuration
 @EnableWebFluxSecurity
 @ConditionalOnProperty(value = "auth.type", havingValue = "LOGIN_FORM")
+@EnableConfigurationProperties(SecurityProperties.class)
 @Slf4j
 public class BasicAuthSecurityConfig extends AbstractAuthSecurityConfig {
+  private static final String NOOP_PASSWORD_PREFIX = "{noop}";
+  private static final Pattern PASSWORD_ALGORITHM_PATTERN = Pattern.compile("^\\{.+}.*$");
 
   @Bean
   public SecurityWebFilterChain configure(ServerHttpSecurity http) {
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/kafbat/ui/config/auth/BasicAuthSecurityConfig.java
```diff
@@ -42,4 +60,42 @@ public SecurityWebFilterChain configure(ServerHttpSecurity http) {
...
+    SecurityProperties.User user = properties.getUser();
+
+    UserDetails userDetails = User.withUsername(user.getName())
+        .password(password(user.getPassword(), passwordEncoder.getIfAvailable()))
+        .roles(StringUtils.toStringArray(user.getRoles()))
+        .build();
+
+    Collection<String> groups = accessControlService.getRoles().stream()
+        .filter(role -> role.getSubjects().stream()
+            .filter(subj -> Provider.BASIC_AUTH.equals(subj.getProvider()))
```
Haarolean, line 76: this looks like an easy implementation, but I don't quite like the idea of having RBAC related business logic in spring security config-related class. We do extract RBAC roles in extractors, I think we should do the same here as well.

api/src/main/java/io/kafbat/ui/config/auth/BasicAuthSecurityConfig.java
```diff
@@ -42,4 +60,42 @@ public SecurityWebFilterChain configure(ServerHttpSecurity http) {
...
+            .filter(subj -> "user".equals(subj.getType()))
+            .anyMatch(subj -> user.getName().equals(subj.getValue()))
+        )
+        .map(Role::getName)
+        .collect(Collectors.toSet());
+
+    return new RbacUserDetailsService(new RbacBasicAuthUser(userDetails, groups));
+  }
+
+  private String password(String password, PasswordEncoder encoder) {
```
Haarolean, line 86: why do we need all of these btw?

Haarolean (changes requested):

api/src/main/java/io/kafbat/ui/model/rbac/provider/Provider.java
```diff
@@ -13,7 +13,9 @@ public enum Provider {
   OAUTH,
 
   LDAP,
-  LDAP_AD;
+  LDAP_AD,
+
+  BASIC_AUTH;
```
wernerdv, line 18: @Haarolean Can I leave the provider name as `BASIC_AUTH` or rename it to `LOGIN_FORM` similar to the value of the `auth.type` parameter?

Haarolean: I think `LOGIN_FORM` is more familiar for the user, as basic auth is only used in the code

api/src/main/java/io/kafbat/ui/config/auth/BasicAuthSecurityConfig.java
Haarolean: I wonder why are _any_ changes to this class are needed? This worked fine before, does introducing something like a user dto breaks something here?

wernerdv: To implement RBAC support for basic auth, it is necessary for `ReactiveUserDetailsService#findByUsername` to return RbacUser (see `AccessControlService#getUser`).
By default, Spring Security creates a MapReactiveUserDetailsService bean for basic auth.
To work with RbacUser, I create a custom RbacUserDetailsService, which requires passing UserDetails to its constructor.

I haven't figured out how to implement this differently yet.
If you have any ideas or if I’ve made a mistake somewhere, please let me know - I’d be happy to fixit.

Haarolean: Ok, gotcha:
- we need a user object with our populated groups/authorities
- default spring impl for login form doesn't support this 
- to override we need a custom `ReactiveUserDetailsService`

api/src/main/java/io/kafbat/ui/config/auth/BasicAuthSecurityConfig.java
```diff
@@ -42,4 +58,39 @@ public SecurityWebFilterChain configure(ServerHttpSecurity http) {
...
+    if (accessControlService.isRbacEnabled()) {
+      RbacBasicAuthAuthoritiesExtractor extractor = new RbacBasicAuthAuthoritiesExtractor(accessControlService);
+
+      return new RbacUserDetailsService(new RbacBasicAuthUser(userDetails, extractor.groups(user.getName())));
+    } else {
+      return new MapReactiveUserDetailsService(userDetails);
+    }
+  }
+
+  private String password(String password, PasswordEncoder encoder) {
```
Haarolean, line 81: I don't think we even need this method, we could delegate this to the encoder which should handle this via `encoder.encode(password);`

api/src/main/java/io/kafbat/ui/config/auth/BasicAuthSecurityConfig.java
```diff
@@ -42,4 +58,39 @@ public SecurityWebFilterChain configure(ServerHttpSecurity http) {
...
+        .password(password(user.getPassword(), passwordEncoder.getIfAvailable()))
+        .roles(StringUtils.toStringArray(user.getRoles()))
+        .build();
+
+    if (accessControlService.isRbacEnabled()) {
+      RbacBasicAuthAuthoritiesExtractor extractor = new RbacBasicAuthAuthoritiesExtractor(accessControlService);
+
+      return new RbacUserDetailsService(new RbacBasicAuthUser(userDetails, extractor.groups(user.getName())));
+    } else {
+      return new MapReactiveUserDetailsService(userDetails);
```
Haarolean, line 77: here we can do a fast return in the beginning of the method

api/src/main/java/io/kafbat/ui/service/rbac/extractor/RbacBasicAuthAuthoritiesExtractor.java
```diff
@@ -0,0 +1,26 @@
...
+  public RbacBasicAuthAuthoritiesExtractor(AccessControlService accessControlService) {
+    this.accessControlService = accessControlService;
+  }
+
+  public Collection<String> groups(String username) {
+    return accessControlService.getRoles().stream()
+        .filter(role -> role.getSubjects().stream()
+            .filter(subj -> Provider.LOGIN_FORM.equals(subj.getProvider()))
+            .filter(subj -> "user".equals(subj.getType()))
+            .anyMatch(subj -> username.equals(subj.getValue()))
```
Haarolean, line 21: i'd suggest `equalsIgnoreCase` here just in case, what do you think?

api/src/test/java/io/kafbat/ui/BasicAuthIntegrationTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+@SpringBootTest
+@ActiveProfiles("rbac-login-form")
+@AutoConfigureWebTestClient(timeout = "60000")
+public class BasicAuthIntegrationTest {
+  @Autowired
+  private WebTestClient client;
+
+  @Test
+  void testUserPermissions() {
```
Haarolean, line 31: thanks for adding a test 😄

Haarolean (dismissed):

wernerdv:
> @wernerdv hey, before you dive in: we thought it looked kinda weird to assign RBAC for a single user, and it seems we could achieve the same result by implementing #344 instead. What do you think?

Implementing a default role makes sense, but that would still require creating a provider for `auth.type=LOGIN_FORM`, or am I mistaken?

germanosin:
> > @wernerdv hey, before you dive in: we thought it looked kinda weird to assign RBAC for a single user, and it seems we could achieve the same result by implementing #344 instead. What do you think?
> 
> Implementing a default role makes sense, but that would still require creating a provider for `auth.type=LOGIN_FORM`, or am I mistaken?

Maybe instead of LoginForm it would be better to add provider wildcard (ANY) and add this role as default?

Haarolean:
> > @wernerdv hey, before you dive in: we thought it looked kinda weird to assign RBAC for a single user, and it seems we could achieve the same result by implementing #344 instead. What do you think?
> 
> Implementing a default role makes sense, but that would still require creating a provider for `auth.type=LOGIN_FORM`, or am I mistaken?

I don't think so, we'd just apply a default role for anyone who can successfully log in, no matter what auth method

wernerdv:
@Haarolean @germanosin Alright, so can we close this PR then?
If you haven't started working on https://github.com/kafbat/kafka-ui/issues/344 yet, I can take on the task.

Haarolean:
> @Haarolean @germanosin Alright, so can we close this PR then? If you haven't started working on #344 yet, I can take on the task.

yeah I think so. Sorry that we thought about it too late :)
Sure, #344 is yours!

PR #960, opened 2025-03-22
Update CVE checks workflow to trigger on changes in 'infra/' directory

### LINKED ISSUE #958: Infra: Limit CVE checks run on main
It shouldn't get triggered if unrelated to the app files are changed, like infra

Brijeshthummar02:

fixes #958 

CVE checks will now run only when infrastructure files change. The workflow won't trigger for unrelated application code changes. Notifications will work correctly for scheduled runs if CVE checks fail.

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

### DIFF
--- a/.github/workflows/cve_checks.yml
+++ b/.github/workflows/cve_checks.yml
@@ -1,9 +1,14 @@
 name: "Infra: CVE checks"
+
 on:
   pull_request:
     types: [ "opened", "reopened", "synchronize" ]
+    paths:
+      - 'infra/**'  # Run only when files in 'infra/' change
   push:
     branches: [ "main" ]
+    paths:
+      - 'infra/**'
   workflow_dispatch:
   schedule:
     # * is a special character in YAML so you have to quote this string
@@ -76,7 +81,7 @@ jobs:
 
   notify:
     needs: check-cves
-    if: ${{ always() && needs.build-and-test.result == 'failure' && github.event_name == 'schedule' }}
+    if: ${{ always() && needs.check-cves.result == 'failure' && github.event_name == 'schedule' }}
     uses: ./.github/workflows/infra_discord_hook.yml
     with:
       message: "Attention! CVE checks run failed! Please fix them CVEs :("


Brijeshthummar02:
@Haarolean what you think, good to go?

Haarolean:
> @Haarolean what you think, good to go?

not quite: It's the other way around - we need to run CVE checks only if the app's code is changed -- be it frontend or backend. So we need to think of list of paths to include here

Brijeshthummar02:
> > @Haarolean what you think, good to go?
> 
> not quite: It's the other way around - we need to run CVE checks only if the app's code is changed -- be it frontend or backend. So we need to think of list of paths to include here

updated it any further changes needed?

Brijeshthummar02:
@Haarolean Follow up on this PR.

.github/workflows/cve_checks.yml
```diff
@@ -2,18 +2,32 @@ name: "Infra: CVE checks"
 on:
   pull_request:
     types: [ "opened", "reopened", "synchronize" ]
+    paths:
+      - 'backend/**'
```
Haarolean, line 6: I feel like there's no need to run these if the code itself been changed but not dependencies. What do you think @Brijeshthummar02 ? 

cc @yeikel

yeikel: My understanding is that this check was designed only for vulnerability scans.

However, there is also a [secret scanning feature](https://trivy.dev/v0.57/docs/scanner/secret/) enabled by default that runs as well.

Based on this, I believe we should not maintain any specific list and instead remove the manual paths configuration altogether, allowing scans to run on any project change, as secrets can theoretically leak in any file.

Alternatively, if we do not intend to use secret scanning, I agree that application-level changes do not need to trigger the CVE scan, and we should consider disabling secret scanning to speed up the CVE scan

Given the value of secret scanning, I suggest we leave it enabled and perform a full project scan as proposed above.

 Alternatively, we can split these into two dedicate steps as needed and maintain the lists. But I think that it ads unnecessary maintenance overhead for very little value. This check runs async and the cost should be low

Haarolean: Github offers secret scanning via other tools as well, not sure we need trivy for that but perhaps that won't hurt.

yeikel: My understanding is that these secret scans are based on heuristics and rules. It is probably a good idea to have both to benefit from the different rules/attempts to discover secrets

PR #1056, opened 2025-05-04
BE: RBAC: Impl default role

seono:

Implement default role
https://github.com/kafbat/kafka-ui/issues/344

I updated each extractor’s extract method to return the defaultRole if it’s present, but I’m not sure if this aligns with the original intention of how extract is supposed to be used.

I tested it using the configuration below
```yaml
rbac:
  default-role:
    name: defaultRole
    permissions:
        - resource: clusterconfig
          actions: [ view ]

        - resource: topic
          value: ".*"
          actions: 
            - VIEW
            - MESSAGES_READ

        - resource: consumer
          value: ".*"
          actions: [ view ]

        - resource: schema
          value: ".*"
          actions: [ view ]

        - resource: connect
          value: ".*"
          actions: [ view ]

        - resource: acl
          actions: [ view ]
  roles: []
```

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

### DIFF
--- a/api/src/main/java/io/kafbat/ui/config/auth/RoleBasedAccessControlProperties.java
+++ b/api/src/main/java/io/kafbat/ui/config/auth/RoleBasedAccessControlProperties.java
@@ -4,6 +4,7 @@ import io.kafbat.ui.model.rbac.Role;
 import jakarta.annotation.PostConstruct;
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.Nullable;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @ConfigurationProperties("rbac")
@@ -11,13 +12,26 @@ public class RoleBasedAccessControlProperties {
 
   private final List<Role> roles = new ArrayList<>();
 
+  private Role defaultRole;
+
   @PostConstruct
   public void init() {
     roles.forEach(Role::validate);
+    if (defaultRole != null) {
+      defaultRole.validateDefaultRole();
+    }
   }
 
   public List<Role> getRoles() {
     return roles;
   }
 
+  public void setDefaultRole(Role defaultRole) {
+    this.defaultRole = defaultRole;
+  }
+
+  @Nullable
+  public Role getDefaultRole() {
+    return defaultRole;
+  }
 }
--- a/api/src/main/java/io/kafbat/ui/model/rbac/Role.java
+++ b/api/src/main/java/io/kafbat/ui/model/rbac/Role.java
@@ -21,4 +21,8 @@ public class Role {
     subjects.forEach(Subject::validate);
   }
 
+  public void validateDefaultRole() {
+    permissions.forEach(Permission::validate);
+    permissions.forEach(Permission::transform);
+  }
 }
--- a/api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
+++ b/api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
@@ -65,21 +65,18 @@ public class AccessControlService {
 
   @PostConstruct
   public void init() {
-    log.info("Initializing Access Control Service");
-    log.info("defaultRole: {}", properties.getDefaultRole());
-    log.info("roles: {}", properties.getRoles());
     if (CollectionUtils.isEmpty(properties.getRoles()) && properties.getDefaultRole() == null) {
       log.trace("No roles provided, disabling RBAC");
       return;
     }
     if (properties.getDefaultRole() != null) {
+      log.trace("Set Default Role Clusters");
       properties.getDefaultRole().setClusters(
           clustersStorage.getKafkaClusters().stream()
             .map(KafkaCluster::getName)
             .collect(Collectors.toList())
       );
     }
-    log.info("defaultRole: {}", properties.getDefaultRole());
     rbacEnabled = true;
 
     if (properties.getDefaultRole() != null) {
--- a/api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
+++ b/api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
@@ -6,6 +6,7 @@ import io.kafbat.ui.config.auth.RoleBasedAccessControlProperties;
 import io.kafbat.ui.model.ClusterDTO;
 import io.kafbat.ui.model.ConnectDTO;
 import io.kafbat.ui.model.InternalTopic;
+import io.kafbat.ui.model.KafkaCluster;
 import io.kafbat.ui.model.rbac.AccessContext;
 import io.kafbat.ui.model.rbac.Permission;
 import io.kafbat.ui.model.rbac.Role;
@@ -14,6 +15,7 @@ import io.kafbat.ui.model.rbac.permission.ConnectAction;
 import io.kafbat.ui.model.rbac.permission.ConsumerGroupAction;
 import io.kafbat.ui.model.rbac.permission.SchemaAction;
 import io.kafbat.ui.model.rbac.permission.TopicAction;
+import io.kafbat.ui.service.ClustersStorage;
 import io.kafbat.ui.service.rbac.extractor.CognitoAuthorityExtractor;
 import io.kafbat.ui.service.rbac.extractor.GithubAuthorityExtractor;
 import io.kafbat.ui.service.rbac.extractor.GoogleAuthorityExtractor;
@@ -53,6 +55,7 @@ public class AccessControlService {
   @Nullable
   private final InMemoryReactiveClientRegistrationRepository clientRegistrationRepository;
   private final RoleBasedAccessControlProperties properties;
+  private final ClustersStorage clustersStorage;
   private final Environment environment;
 
   @Getter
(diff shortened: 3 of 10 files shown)


api/src/main/java/io/kafbat/ui/config/auth/RoleBasedAccessControlProperties.java
```diff
@@ -4,20 +4,34 @@
 import jakarta.annotation.PostConstruct;
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.Nullable;
```
germanosin: Let's use jakarta

api/src/main/java/io/kafbat/ui/model/rbac/Role.java
```diff
@@ -22,4 +22,9 @@ public void validate() {
     subjects.forEach(Subject::validate);
   }
 
+  // default role need only permissions
```
germanosin: Let's create a separate class for it not to mix logic

api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
```diff
@@ -62,13 +65,31 @@ public class AccessControlService {
 
   @PostConstruct
   public void init() {
-    if (CollectionUtils.isEmpty(properties.getRoles())) {
+    if (CollectionUtils.isEmpty(properties.getRoles()) && properties.getDefaultRole() == null) {
       log.trace("No roles provided, disabling RBAC");
       return;
     }
+    if (properties.getDefaultRole() != null) {
```
germanosin: I think it's better to handle in another way, instead of adding all clusters just apply it to any cluster

api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
```diff
@@ -62,13 +65,31 @@ public class AccessControlService {
...
+      properties.getDefaultRole().setClusters(
+          clustersStorage.getKafkaClusters().stream()
+            .map(KafkaCluster::getName)
+            .collect(Collectors.toList())
+      );
+    }
     rbacEnabled = true;
 
-    this.oauthExtractors = properties.getRoles()
+    if (properties.getDefaultRole() != null) {
```
germanosin: This also looks redundant. Default role should be applied to any subject

api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
```diff
@@ -132,6 +162,9 @@ public static Mono<AuthenticatedUser> getUser() {
 
   private boolean isClusterAccessible(String clusterName, AuthenticatedUser user) {
     Assert.isTrue(StringUtils.isNotEmpty(clusterName), "cluster value is empty");
+    if (properties.getDefaultRole() != null) {
```
germanosin: It looks a bit strange. Why do we need this?

api/src/main/java/io/kafbat/ui/service/rbac/extractor/CognitoAuthorityExtractor.java
```diff
@@ -39,8 +39,13 @@ public Mono<Set<String>> extract(AccessControlService acs, Object value, Map<Str
 
     var usernameRoles = extractUsernameRoles(acs, principal);
     var groupRoles = extractGroupRoles(acs, principal);
+    var unionRoles = Sets.union(usernameRoles, groupRoles);
 
-    return Mono.just(Sets.union(usernameRoles, groupRoles));
+    if (unionRoles.isEmpty() && acs.getDefaultRole() != null) {
```
germanosin: Why should we change this?

api/src/main/java/io/kafbat/ui/service/rbac/extractor/OauthAuthorityExtractor.java
```diff
@@ -43,8 +43,13 @@ public Mono<Set<String>> extract(AccessControlService acs, Object value, Map<Str
 
     var usernameRoles = extractUsernameRoles(acs, principal);
     var roles = extractRoles(acs, principal, additionalParams);
+    var unionRoles = Sets.union(usernameRoles, roles);
 
-    return Mono.just(Sets.union(usernameRoles, roles));
```
germanosin: I would prefer not to change anything inside extractors

api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
```diff
@@ -132,10 +143,15 @@ public static Mono<AuthenticatedUser> getUser() {
 
   private boolean isClusterAccessible(String clusterName, AuthenticatedUser user) {
```
seono, line 143: @germanosin 
Thanks for the feedback! 
Applied most changes. Updated this method to fallback to defaultRole 
when checking cluster accessibility instead of relying solely on properties.getRoles().

seono (commented):
@germanosin 
Thanks for the review.
I updated the code to work with the RBAC definition below.

```yaml
...

rbac:
  default-role:
    name: read-only
    clusters:
      - dev-cluster
    permissions:
      - resource: clusterconfig
        actions: [ "view" ]

      - resource: topic
        value: ".*"
        actions: 
          - VIEW
          - MESSAGES_READ

      - resource: consumer
        value: ".*"
        actions: [ view ]

      - resource: schema
        value: ".*"
        actions: [ view ]

      - resource: connect
        value: ".*"
        actions: [ view ]

      - resource: acl
        actions: [ view ]

  roles:
    - name: admin
      clusters:
        - dev-cluster
        - prod-cluster
      subjects:
        - provider: oauth_github
          type: user
          value: "seono"
...
```

api/src/main/java/io/kafbat/ui/model/rbac/DefaultRole.java
```diff
@@ -0,0 +1,21 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.util.ArrayList;
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class DefaultRole {
+
+  private String name;
```
germanosin: why do we need name?

api/src/main/java/io/kafbat/ui/model/rbac/DefaultRole.java
```diff
@@ -0,0 +1,21 @@
...
+  private List<String> clusters;
```
germanosin: According to your code you apply it to any cluster, why do we need them here?

api/src/main/java/io/kafbat/ui/model/rbac/DefaultRole.java
```diff
@@ -0,0 +1,21 @@
...
+  private List<Permission> permissions = new ArrayList<>();
+
+  public void validate() {
+    checkArgument(clusters != null && !clusters.isEmpty(), "Default role clusters cannot be empty");
```
germanosin: Why I should set list of clusters for default role?

germanosin (changes requested):

api/src/main/java/io/kafbat/ui/model/rbac/DefaultRole.java
seono: Looks like I brought it together with Role when separating things. After checking, it seems unnecessary — I’ll go ahead and remove it

api/src/main/java/io/kafbat/ui/model/rbac/DefaultRole.java
seono: I intended to specify the cluster even for the DefaultRole.
So, I updated the AccessControlService.isClusterAccessible() method to also check whether the DefaultRole includes a matching cluster when falling back.
I thought there could be cases where the DefaultRole should apply only to cluster A, and not to others.

seono: I removed the clusters field from the DefaultRole class and updated the validate method, as well as the relevant code in AccessControlService and AuthorizationControl

germanosin: Great! Thanks for a quick fix

api/src/main/java/io/kafbat/ui/config/auth/RoleBasedAccessControlProperties.java
```diff
@@ -11,13 +13,26 @@ public class RoleBasedAccessControlProperties {
...
+    if (defaultRole != null) {
+      defaultRole.validate();
+    }
   }
 
   public List<Role> getRoles() {
     return roles;
   }
 
+  public void setDefaultRole(DefaultRole defaultRole) {
```
Haarolean, line 30: Perhaps an explicit setter is redundant here

api/src/test/java/io/kafbat/ui/service/rbac/AccessControlServiceDefaultRoleRbacEnabledTest.java
```diff
@@ -0,0 +1,102 @@
...
+      // Mock static method to get security context
+      ctxHolder.when(ReactiveSecurityContextHolder::getContext).thenReturn(Mono.just(securityContext));
+      runnable.run();
+    }
+  }
+
+  @Test
+  void validateAccess() {
+    withSecurityContext(() -> {
+      when(user.groups()).thenReturn(List.of(DEFAULT_ROLE));
```
Haarolean, line 81: Not following here, what are we testing here for? Given that we mock users' groups, it seems that no newly implemented logic is gonna be tested against?

api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
```diff
@@ -86,7 +87,7 @@ public void init() {
         .flatMap(Set::stream)
         .collect(Collectors.toSet());
 
-    if (!properties.getRoles().isEmpty()
+    if (!(properties.getRoles().isEmpty() && properties.getDefaultRole() == null)
```
Haarolean: While we're at it, can we refactor this monstrosity a bit? A negative check on a compound `and` plus a few other `and`s takes a lot of cognitive load to parse.

Haarolean (changes requested):

api/src/test/java/io/kafbat/ui/service/rbac/AccessControlServiceDefaultRoleRbacEnabledTest.java
seono: @Haarolean 
The purpose of this test is to verify that, regardless of the cluster, the given context is correctly mapped and allowed based only on the permissions defined for the mock user with the Default role.

api/src/main/java/io/kafbat/ui/config/auth/RoleBasedAccessControlProperties.java
seono: The defaultRole field is nullable, so it can’t be declared as final, which means a setter is required for property binding.
If the setter is removed, it results in a binding error like the one below:

```
Failed to bind properties under 'rbac' to io.kafbat.ui.config.auth.RoleBasedAccessControlProperties:  
Reason: java.lang.IllegalStateException: No setter found for property: default-role
```

api/src/main/java/io/kafbat/ui/service/rbac/AccessControlService.java
seono: I’ve refactored the condition into a separate variable (hasRolesConfigured) so that the logic is easier to read and understand.

api/src/main/java/io/kafbat/ui/config/auth/RoleBasedAccessControlProperties.java
Haarolean: Indeed! Thanks

api/src/test/java/io/kafbat/ui/service/rbac/AccessControlServiceDefaultRoleRbacEnabledTest.java
Haarolean: Ah, I see. And the other test is actually the one which validates that the user has an access despite having no explicitly defined roles.

Haarolean (approved):

Haarolean:
@seono, thanks for your first contribution to Kafbat UI!

PR #1108, opened 2025-05-29
Allow disabling GitHub release info

stklcode:

The application info object contains details about the latest version which is queried online from GitHub API. While this info can be useful, it raises warning when Kafbat UI is deployed in environments with restricted internet access.

Introduce a new property `github.release.info.enabled` (default: `true`) that can be set to `false` to disable release info and just display the currently running version/commit.

The original feature was introduced in https://github.com/provectus/kafka-ui/pull/3570

Apparently, the issue was already known during development (https://github.com/provectus/kafka-ui/pull/3570#pullrequestreview-1361235598):
> We need a boolean property to make it possible to opt-out of such checks in air-gap environments.

Did I miss anything or was this point just ignored? Anyway with this PR now there it is.

- [ ] No need to
  Start app with environment variable or property, e.g. `GITHUB_RELEASE_INFO_ENABLED=false`
- [x] Unit checks
- [ ] Integration checks

  Currently existing timeout property is not documented anywhere ... Should we add both at some place?


api/src/main/java/io/kafbat/ui/service/ApplicationInfoService.java
```diff
@@ -44,36 +45,49 @@ public ApplicationInfoService(DynamicConfigOperations dynamicConfigOperations,
...
                                 @Value("${" + GITHUB_RELEASE_INFO_TIMEOUT + ":10}") int githubApiMaxWaitTime) {
     this.applicationContext = applicationContext;
     this.dynamicConfigOperations = dynamicConfigOperations;
     this.buildProperties = Optional.ofNullable(buildProperties).orElse(new BuildProperties(new Properties()));
     this.gitProperties = Optional.ofNullable(gitProperties).orElse(new GitProperties(new Properties()));
-    githubReleaseInfo = new GithubReleaseInfo(githubApiMaxWaitTime);
+    if (githubInfoEnabled) {
+      this.githubReleaseInfo = new GithubReleaseInfo(githubApiMaxWaitTime);
+    } else {
+      this.githubReleaseInfo = null;
```
Haarolean, line 57: let's either mark the field as nullable or replace it with some empty object instead to prevent NPEs

Haarolean: Let's also output a bold warning into logs on startup in case this is disabled that running outdated versions is dangerous due to CVEs and we will **_not_** provide support for outdated versions per our security policy

api/src/main/java/io/kafbat/ui/service/ApplicationInfoService.java
```diff
@@ -44,36 +45,49 @@ public ApplicationInfoService(DynamicConfigOperations dynamicConfigOperations,
...
+    var releaseInfo = githubReleaseInfo != null ? githubReleaseInfo.get() : null;
     return new ApplicationInfoDTO()
         .build(getBuildInfo(releaseInfo))
         .enabledFeatures(getEnabledFeatures())
         .latestRelease(convert(releaseInfo));
   }
 
   private ApplicationInfoLatestReleaseDTO convert(GithubReleaseInfo.GithubReleaseDto releaseInfo) {
+    if (releaseInfo == null) {
+      return null;
```
Haarolean, line 71: `@Nullable` for the method please, or a non-null approach

api/src/main/java/io/kafbat/ui/util/GithubReleaseInfo.java
```diff
@@ -8,6 +8,7 @@
 
 @Slf4j
 public class GithubReleaseInfo {
+  public static final String GITHUB_RELEASE_INFO_ENABLED = "github.release.info.enabled";
```
Haarolean, line 11: Please document this property in the docs too: https://github.com/kafbat/ui-docs

frontend/src/components/Version/Version.tsx
```diff
@@ -19,7 +19,7 @@ const Version: React.FC = () => {
 
   return (
     <S.Wrapper>
-      {!isLatestRelease && (
+      {isLatestRelease === false && (
```
Haarolean, line 22: I'd like to see a visual difference between "no latest release present" and "user disabled version check" in the UI. This is crucial as it's often only the UI screenshots I have to deduct things from.

Haarolean (changes requested):
Hi, thanks for the PR. I've left some inline comments, PTAL!

frontend/src/components/Version/Version.tsx
stklcode: Any suggestion how such information should look?

Frankly, I don't see any benefit in such information. Currently running version is displayed, if the check failed there's a warning symbol and the information if there's an update available or not can change 1 second after a screenshot was taken, because it's dynamic info.

api/src/main/java/io/kafbat/ui/service/ApplicationInfoService.java
stklcode: > Let's also output a bold warning into logs on startup ...

A real "warning", i.e. WARN log level? INFO ... OK, but warning seems little overdone. I'm running about 400 instances in various test clusters and remote systems, no online access, updates and security scans are done by external tooling. Simply shifting Firewall/eBPF warnings to application warnings was not what I had in mind....

frontend/src/components/Version/Version.tsx
stklcode: A red warning, really?

I've learned to ignore the yellow icon due do failing version checks and yet there are frequent questions if there's something wrong. So annoying users, who have no control whatsoever about the deployed version in our case, with read warning sign that typically means "something is wrong here" ...

Our use case are - you guessed it - isolated systems, mostly dev/staging environments. We allow users some access to the underlying Kafka and deployment/updates are done by some automation/opt team who don't see the UI warnings anyway.

We were looking for a solution to remove the frontend warning, not making it look worse.

api/src/main/java/io/kafbat/ui/service/ApplicationInfoService.java
Haarolean: Any should work fine, there's no guideline on this rn, feel free to use any of these

api/src/main/java/io/kafbat/ui/service/ApplicationInfoService.java
stklcode: I get the point. But does a warning really prevent these folks from requesting anything? My experience is you can write "UNMAINTEINED" in 50px bold at the top of a GH issue template and you still get new tickets with the warning not even removed :man_shrugging:

frontend/src/components/Version/Version.tsx
stklcode: On the oder end, I've seen questions from users about the warning of "UNKNOWN" version and regular tickets from ops about log warnings that nobody can do anything about.

We have filters on our network logs to prevent alerting from unauthorized access attempts and injected a logging config to suppress the "Authentication is disabled" warning for instances running behind an external auth gateway. But we cannot filter the human eye and regular "just ignore the warning" is not really a satisfactory answer.

api/src/main/java/io/kafbat/ui/util/GithubReleaseInfo.java
stklcode: Done: https://github.com/kafbat/ui-docs/pull/56

Also added the 2 pre-existing properties for this feature (timeout, update-rate)

api/src/main/java/io/kafbat/ui/service/ApplicationInfoService.java
Haarolean: >yet another one to ignore

do you export them to some external systems like sentry or something?

api/src/test/java/io/kafbat/ui/service/ApplicationInfoServiceTest.java
```diff
@@ -1,17 +1,43 @@
...
+  @Autowired
+  private DynamicConfigOperations dynamicConfigOperations;
+
   @Test
   void testCustomGithubReleaseInfoTimeout() {
     assertEquals(100, service.githubReleaseInfo().getGithubApiMaxWaitTime());
   }
+
+  @Test
+  void testDisabledReleaseInfo() {
```
Haarolean, line 25: Thank you for adding the tests

frontend/src/components/Version/Version.tsx
Haarolean: I think we could hide the whole icon tbh. Even if the person reporting issues is not the same as the one managing the installation, they can still check if the version is up to date by clicking the commit hash. Given the counterpoint by @stklcode about "users screaming due to an alert", it might be alright.
@germanosin thoughts?

Haarolean (approved):

api/src/main/java/io/kafbat/ui/service/ApplicationInfoService.java
stklcode: Yes, quite a few deployments on K8s where pod's logs are scraped and some generic checks for warnings and errors.

Not really a big deal.
Mostly dev/qa stuff where this is non-critical and for the remaining staging/prod-like namespaces there are filters in the deployment templates - almost impossible to run every third-party tool 100% warning-free in every environment, but normal under normal conditions we don't want any warning without actual impact. So just accept some "unfixable" ones (like version checks in an offline environment) explicitly and move on.

frontend/src/components/Version/Version.tsx
stklcode: Any update on this one?

I'd be happy to remove the latest commit again (hide the icon) if we agree on that or keep is as is.
Just rebased the code and resolved a minor merge conflict.

Haarolean: @stklcode ok let's hide the icon (based on a property ofc)

stklcode: OK. I removed the latest commit, so now the icon is gone if `github.release.info.enabled=false`.

I will extract a minor change (warning icon color from theme) into a separate PR, as it is now little out of scope here. (#1282)

stklcode:
Any news on that one @Haarolean @germanosin ?

I removed  the warning icon as requested 2 months ago.
Updating the branch from time to time to keep it in a mergeable state without seeing any progress is becoming little tedious...

PR #1143, opened 2025-06-21
BE: Chore: Bump Spring Boot to 3.5.3

yeikel:
**What changes did you make? (Give an overview)**

- Upgraded Spring Boot to the latest version (`3.5.3`) to fetch the latest bug fixes and features as well as fixing CVE-2025-41234 
- Removed several dependency overrides that are no longer necessary since they were upgraded by Spring
- Removed the explicit inclusion of `Junit5` , as it is provided transitively by `spring-boot-starter-test`. Keeping it explicitly declared leads to version conflicts, which goes against the [JUnit project's recommendations for Spring Boot](https://docs.junit.org/snapshot/user-guide/#running-tests-build-spring-boot).

image

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -53,7 +53,7 @@ dependencies {
         exclude group: 'io.projectreactor.ipc', module: 'reactor-netty'
     }
 
-    runtimeOnly(libs.micrometer.registry.prometheus){
+    runtimeOnly(libs.micrometer.registry.prometheus) {
         exclude group: 'com.google.protobuf', module: 'protobuf-java' because("Micrometer uses protobuf-java 4.x, which is incompatible with protobuf-java 3.x used by various dependencies of this project. See https://github.com/prometheus/client_java/issues/1431")
     }
 
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -53,15 +53,13 @@ dependencies {
         exclude group: 'io.projectreactor.ipc', module: 'reactor-netty'
     }
 
-    runtimeOnly libs.micrometer.registry.prometheus
+    runtimeOnly(libs.micrometer.registry.prometheus){
+        exclude group: 'com.google.protobuf', module: 'protobuf-java' because("Micrometer uses protobuf-java 4.x, which is incompatible with protobuf-java 3.x used by various dependencies of this project. See https://github.com/prometheus/client_java/issues/1431")
+    }
 
     // CVE Fixes
     implementation libs.apache.commons.compress
     implementation libs.okhttp3.logging.intercepter
-    implementation libs.json.smart
-    implementation libs.netty.common
-    implementation libs.netty.handler
-    implementation libs.spring.context
 
     implementation libs.modelcontextprotocol.spring.webflux
     implementation libs.victools.jsonschema.generator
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -21,7 +21,6 @@ lombok = '1.18.34'
 odd-oddrn-generator = '0.1.17'
 odd-oddrn-client = '0.1.41'
 cel = '0.3.0'
-junit = '5.11.2'
 mockito = '5.16.0'
```
yeikel, line 27: JUnit 5, in particular, was causing a direct conflict that might be resolved by upgrading or aligning versions. However, since it’s brought in transitively by  one our direct test dependencies, maintaining version alignment doesn't seem worth pursuing.

Mockit and assertj are also included transitively and technically falls into the same category, but it didn’t cause any issues during this upgrade.

Overall, reducing the number of explicitly managed dependencies isn’t a bad idea, so I’m open to removing them as well.


@Haarolean What do you think?

api/build.gradle
```diff
@@ -53,15 +53,13 @@ dependencies {
         exclude group: 'io.projectreactor.ipc', module: 'reactor-netty'
     }
 
-    runtimeOnly libs.micrometer.registry.prometheus
+    runtimeOnly(libs.micrometer.registry.prometheus){
+        exclude group: 'com.google.protobuf', module: 'protobuf-java' because("Micrometer uses protobuf-java 4.x, which is incompatible with protobuf-java 3.x used by various dependencies of this project. See https://github.com/prometheus/client_java/issues/1431")
+    }
```
yeikel: This may cause issues for anyone using this functionality, but there does not seem to be a clear fix for this other than  https://github.com/prometheus/client_java/issues/1431

None of our tests seem to be flagging this right now but it may be a matter of coverage

@Haarolean What do you think?

yeikel: For now, I pushed b57acf5c60741baa8cf195dacd9693bfc9bc2107 which we can revert as needed

Haarolean: Yeah I doubt we have tests for this. However we should have a docker-compose somewhere in documentation/compose to try it out with prometheus.

gradle/libs.versions.toml
Haarolean: I don't think it's a good idea to use some random junit/mockito versions pulled transitively. Bumping some other unrelated dependency then can break things, and junit is kinda known for breaking things sometimes.

yeikel: Do you suggest that we continue to sync them? Because what motivated this change is to align them with what Spring boot Uses as it was failing due to a conflict

Haarolean: I'd align them (explicitly) with whatever spring boot uses itself, so in this case we'd have the same issues like we have now (version mismatch) rather than suddenly failing tests (say, when spring boot is bumped next time). What do you think?

api/build.gradle
germanosin: No problem, we don’t rely on the protobuf export format.

gradle/libs.versions.toml
germanosin: Agree with @Haarolean, let's align it explicitly

yeikel: Thanks for the suggestion. I reverted these changes

Haarolean:
@yeikel thank you!

PR #1153, opened 2025-06-26
Add all changes to connect with GCP Schema Registries

magnusdriver:

Added new cluster property "gcpSchemaRegistry" to enable compatibility with new [GCP Schema Registries](https://cloud.google.com/managed-service-for-apache-kafka/docs/schema-registry/schema-registry-overview).
This property can be used in config.yaml file like this:

```yaml
kafka:
  clusters:
    - name: local  # Unique name identifier for the Kafka cluster
      bootstrap-servers: kafka1:9092,kafka2:9092  # List of Kafka broker addresses

      schemaRegistry: https://managedkafka.googleapis.com/v1/projects/PROJECT_ID/locations/REGION/schemaRegistries/REGISTRY_ID
      gcpSchemaRegistry: true # false by default
```

Or can be enabled with the env variable: `KAFKA_CLUSTERS_0_GCPSCHAMAREGISTRY=true`

Important changes in code are:

- Add bearer token generation for auth with GCP Schema Registry in WebClientConfigurator
- Add custom bearer token provider in SchemaRegistrySerde
- Add the new `gcpSchemaRegistry` config property in kafbat-ui-api
- Add new schema fields kafka-sr-api for compatibility level as GCP Schema Registry return different format.
- Add condition to use compatibility or compatibliityLevel field based on gcpSchemaRegistry enabled or not in
  SchemaRegistryService

It's needed to use `gcloud auth application-default login` to get the credentials to connect to GCP Schema 
Registries or run kafka-ui in a compute engine instance or GKE with a service account with the needed permissions.

Note: This functionality only works now with Avro schema types as these are the ones I work with.
It could work with Protobuf schemas, but not with JSON ones as GCP Schema Registries don't 
support them.

  Changes tested connecting to GCP and Confluent Schema Registries to check:
  
  - All Schemas registered are shown in Schema Registry dashboard 
  - Topic events are deserialized correctly in the topics dashboards.
  
  It was impossible to pass unit tests even before adding any change :'(

### DIFF
--- a/api/src/main/java/io/kafbat/ui/serdes/builtin/sr/MessageFormatter.java
+++ b/api/src/main/java/io/kafbat/ui/serdes/builtin/sr/MessageFormatter.java
@@ -11,6 +11,7 @@ import io.confluent.kafka.serializers.KafkaAvroDeserializerConfig;
 import io.confluent.kafka.serializers.json.KafkaJsonSchemaDeserializer;
 import io.confluent.kafka.serializers.protobuf.KafkaProtobufDeserializer;
 import io.kafbat.ui.util.jsonschema.JsonAvroConversion;
+import java.util.HashMap;
 import java.util.Map;
 import lombok.SneakyThrows;
 
@@ -18,9 +19,10 @@ interface MessageFormatter {
 
   String format(String topic, byte[] value);
 
-  static Map<SchemaType, MessageFormatter> createMap(SchemaRegistryClient schemaRegistryClient) {
+  static Map<SchemaType, MessageFormatter> createMap(SchemaRegistryClient schemaRegistryClient,
+                                                     boolean gcpSchemaRegistry) {
     return Map.of(
-        SchemaType.AVRO, new AvroMessageFormatter(schemaRegistryClient),
+        SchemaType.AVRO, new AvroMessageFormatter(schemaRegistryClient, gcpSchemaRegistry),
         SchemaType.JSON, new JsonSchemaMessageFormatter(schemaRegistryClient),
         SchemaType.PROTOBUF, new ProtobufMessageFormatter(schemaRegistryClient)
     );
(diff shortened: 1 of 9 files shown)


api/src/main/java/io/kafbat/ui/serdes/builtin/sr/MessageFormatter.java
```diff
@@ -20,9 +20,10 @@ interface MessageFormatter {
   String format(String topic, byte[] value);
 
   static Map<SchemaType, MessageFormatter> createMap(SchemaRegistryClient schemaRegistryClient,
-                                                     boolean gcpSchemaRegistry) {
+                                                     //boolean gcpSchemaRegistry) {
```
germanosin: Please remove commented code

magnusdriver:
I finally exposed the BEARER_AUTH_CUSTOM_PROVIDER_CLASS parameter. I didn't need the bearer scope for this use case. If you think it's still  interesting to expose that parameter I'll add it. Maybe pass it to WebClientConfigurator.configureBearerTokenAuth as parameter for future use cases?

api/src/main/java/io/kafbat/ui/util/WebClientConfigurator.java
```diff
@@ -24,11 +28,18 @@
...
+import org.springframework.web.reactive.function.client.ClientRequest;
+import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
 import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Mono;
 import reactor.netty.http.client.HttpClient;
 
+
 public class WebClientConfigurator {
 
+  private static final String GCP_BEARER_AUTH_CUSTOM_PROVIDER_CLASS =
```
Haarolean: `WebClientConfigurator` is a generic helper/utility class to build `WebClient` instances. It should not be exposed to any specific/concrete implementations of unrelated features, like gcp.

contract/src/main/resources/swagger/kafka-sr-api.yaml
```diff
@@ -381,8 +381,13 @@ components:
             properties:
                 compatibilityLevel:
                     $ref: '#/components/schemas/Compatibility'
-            required:
```
Haarolean, line 384: Why has compatibility become no longer required?

api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
```diff
@@ -166,6 +178,11 @@ private static SchemaRegistryClient createSchemaRegistryClient(List<String> urls
           keyStorePassword);
     }
 
+    if (bearerAuthCustomProviderClass != null) {
```
Haarolean, line 181: Why is this called a `class`? The implementation rather retrieves a token via custom implementation from GCP which is later used as a bearer token value, there are no "classes" in use per se.

contract/src/main/resources/swagger/kafka-sr-api.yaml
magnusdriver: I'm trying to change it

magnusdriver: I finally added a new compatibilityConfigGcp schema detached from the existent compatibilityConfig. And I had to modify code in SchemaRegistryService class. 

If this new modification is not ok then please let me know if you have any recommendation about how to address this as the only alternative I can think of is to add a new kafka-gcp-sr-api and a new GcpSchemaRegistryService class.

api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
magnusdriver: I use this to reference this [class](https://github.com/googleapis/managedkafka/blob/main/kafka-java-auth/src/main/java/com/google/cloud/hosted/kafka/auth/GcpBearerAuthCredentialProvider.java). And I changed the bearer token implementation for the WebClient using this class too. Sorry 🙏 .

contract/src/main/resources/swagger/kafka-sr-api.yaml
Haarolean: Not sure it's the best solution here:
- this breaks backward compatibility
- every other new value (AwsCompatibility, AzureCompatibility, etc) will expand this list and break the backward compatibility further

@germanosin what do you think?

Haarolean (changes requested):

contract/src/main/resources/swagger/kafka-sr-api.yaml
```diff
@@ -176,7 +176,9 @@ paths:
                     content:
                         application/json:
                             schema:
-                                $ref: '#/components/schemas/CompatibilityConfig'
+                                oneOf:
```
germanosin: This approach seems messy. If GCP SR uses a different API, we should define a separate contract for it.

magnusdriver: Sorry for taking so long. But I finally managed to decouple both APIs. To make it work I decided to create a GcpSchemaRegistryService and a GcpKafkaSrMapper and make use of them in the SchemaController based on a KafkaCluster variable called "isGcpSchemaRegistry" (bool) initialized in the KafkaClusterFactory. I think with this approach would be easy to undo this changes if Google decides at some point to make their SchemaRegistry fully compliant with the standard SR API.

The Serdes remains as it was because the problematic "compatibility" field is not used there. For that component just the custom auth class is needed.

Issue #1088, opened 2025-05-22
Open Kafka Connect connectors in a separate tab


Gleb-Shipilov:
### Issue submitter TODO list

- [x] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [x] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md) and the feature is not present there

### Describe the feature you're interested in

**Prerequisites**
Currently it's impossible to open Kafka Connect connectors in a new tab. Yet it's possible for topics, consumers and Schema Registry schemas

**To do**
Add possibility to open Kafka Connect connectors in a new tab

**Acceptance criteria**
Given: a user is on a web page with connectors
When: a user makes right button click (CMD + Click / CTRL + Click)
Then: connector details are opened in a separate tab

### Version you're running

2ce0662

PR #1165, opened 2025-07-04
UX: Open KC connectors in a separate tab

Nilumilak:
Resolves: #1088


Vixtir (approved):

fallen-up:
- now all four initial columns are shown in bold. is it ok?
- the context menu appears for all elements in the first four columns (Name, Connect, Type, Plugin), not just Name - which might confuse users, imho. for example, if someone clicks next to a plugin name, it would make more sense to open a page filtered by that plugin name. same for type and connect.

PR #1181, opened 2025-07-13
Infra: Create dedicated Dependabot group for Spring dependencies

yeikel:
## What changes did you make? (Give an overview)

While Dependabot's grouped updates can be convenient, they should be used with care, particularly in projects with a large number of dependencies like this one. 

Grouped updates often combine changes that are not compatible with each other, which can make upgrades more difficult and increase the risk of breaking the build. For instance, https://github.com/kafbat/kafka-ui/pull/1163 includes over 20 updates, and it's not surprising that the build is currently failing. Reviewing and validating that many changes at once is a significant challenge.

This pull request takes an initial step toward organizing updates into smaller, more focused groups, beginning with Spring . Spring was selected because it is both a common source of security advisories and a foundational component of the project’s that should be updated regularly.

Ultimately, we are using grouped dependencies for convenience, but it should not cause us to miss critical upgrades due to the nature and complexity of the groups as the project's complexity grows 

- [x] Manually merging this pull request into my fork

It created this pull request: https://github.com/yeikel/kafka-ui/pull/197/files

6a010535647bf3970b01156ff51666970c

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,15 @@ updates:
     - "type/dependencies"
     - "scope/backend"
   groups:
-    gradle-dependencies:
+    spring-boot-dependencies:
+      patterns:
+        - "org.springframework.boot:*"
+        - "io.spring.dependency-management"
+      # We will handle major upgrades manually
+      update-types:
+        - "patch"
+        - "minor"
+    other-dependencies:
       patterns:
         - "*"
       update-types:


.github/dependabot.yml
```diff
@@ -14,7 +14,15 @@ updates:
...
   groups:
-    gradle-dependencies:
+    spring-boot-dependencies:
+      patterns:
+        - "org.springframework.boot:*"
+        - "io.spring.dependency-management"
+      # We will handle major upgrades manually
+      update-types:
+        - "patch"
+        - "minor"
```
yeikel, line 22: I copied and documented this from the other entries but I don't agree necessarily. 

Maybe we should have another group for major?

Haarolean:
Yeah, so the idea why I've grouped all (!) the dependency bumps in a single group is to reduce dependabot's PR spam. I thought we would've excluded the dependencies causing build/test failures from this chonky PRs, but neither we had much time addressing single-dependency PRs before, nor we do have for this chonks now. 
Why this supposed to work fine with security updates is that dependabot creates separate PRs for security bumps anyway, that's why it wasn't considered a problem.

I agree with the approach overall, let's start creating some dedicated groups.

.github/dependabot.yml
Haarolean: Maybe it makes sense to indeed have that group (via separate PRs) so we'll at least know there's a new version out quicker

yeikel: Will do. I'll raise a separate PR after this is merged

yeikel:
> Yeah, so the idea why I've grouped all (!) the dependency bumps in a single group is to reduce dependabot's PR spam. I thought we would've excluded the dependencies causing build/test failures from this chonky PRs, but neither we had much time addressing single-dependency PRs before, nor we do have for this chonks now. Why this supposed to work fine with security updates is that dependabot creates separate PRs for security bumps anyway, that's why it wasn't considered a problem.
> 
> I agree with the approach overall, let's start creating some dedicated groups.

Thank you for the explanation. I understand that dependabot pull requests can overwhelm the maintainers for such a large project like this.  Let's start small with two groups for now and we can iterate as we see fit 

Thank you again!

.github/dependabot.yml
```diff
@@ -14,7 +14,15 @@ updates:
     - "type/dependencies"
     - "scope/backend"
   groups:
-    gradle-dependencies:
+    spring-boot-dependencies:
+      patterns:
```
Haarolean, line 16: We've got some non-boot stuff too, like `spring-security-ldap` and `spring-security-oauth2-resource-server` from `libs.versions.toml`, don't we need to include packages of these two here as well?

yeikel: This should not be needed unless we decide to define them separately in the future. 

Right now, we are aligning them with the same version as Spring boot

```properties
spring-starter-security = { module = 'org.springframework.boot:spring-boot-starter-security', version.ref = 'spring-boot' }
```

yeikel:
@Haarolean What blocks merging this one? I have at least 2 follow up PRs planned after this one

Thanks again!

Issue #1248, opened 2025-08-06
FE: Auth: Logout button should consider context path


romanfurst:
### Issue submitter TODO list

- [x] I've looked up my issue in [FAQ](https://ui.docs.kafbat.io/faq/common-problems)
- [x] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [x] I've tried running `main`-labeled docker image and the issue still persists there
- [x] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md)

### Describe the bug (actual behavior)

I believe there is little misbehaviour in logout button behaviour. 
If context path is configured e.g. `SERVER_SERVLET_CONTEXT_PATH: /tt-test01` , kafbat-ui correctly listens on this path and everything works smoothly.
However, a small trouble occurs when trying to logout. After hiting the logout button, the action is triggered on `/logout` instead of ` /tt-test01/logout`,  which result in 404,  as shown in below screenshots.
As auth method oauth-keycloak was used. 

<img width="1697" height="102" alt="Image" src="https://github.com/user-attachments/assets/8e0787c1-944b-4b3e-a7e4-ea841d99aafb" />

<img width="1697" height="138" alt="Image" src="https://github.com/user-attachments/assets/7a07e36a-cdfd-4b24-a7e7-794745775fe1" />

<img width="1709" height="130" alt="Image" src="https://github.com/user-attachments/assets/621fe974-8640-41a2-bc2e-b877214146c1" />

### Expected behavior

logout button action should use correct context path as shown below:

<img width="1356" height="213" alt="Image" src="https://github.com/user-attachments/assets/3fc09246-c621-4a40-9d19-39f5240f9ca6" />

### Your installation details

1) helm chart version: 1.3.0
2) kafbat-ui versio: 1.5.1

3) important setting sections:

`SERVER_SERVLET_CONTEXT_PATH: /tt-test01`

and 

```
    auth:
      type: OAUTH2
      oauth2:
        client:
          keycloak:
            clientId: foo-bar-client
            clientSecret: *********
            scope: 'openid'
            issuer-uri: https://iam.foobar.com.com/auth/realms/myrelam
            user-name-attribute: preferred_username
            client-name: keycloak
            provider: keycloak
            custom-params:
              type: oauth
```

### Steps to reproduce

install kafba-ui with SERVER_SERVLET_CONTEXT_PATH set. Login and then logout.

[...]

PR #1258, opened 2025-08-11
FE: fixes(#1248): context path for logout

pandeysambhi:

fixes - #1248

Considered context path for auth log out

- [ ] No need to
- [x] Unit checks
- [ ] Integration checks

### DIFF
--- a/frontend/src/components/NavBar/UserInfo/UserInfo.tsx
+++ b/frontend/src/components/NavBar/UserInfo/UserInfo.tsx
@@ -19,7 +19,7 @@ const UserInfo = () => {
         </S.Wrapper>
       }
     >
-      <DropdownItem href="/logout">Log out</DropdownItem>
+      <DropdownItem href={`${window.basePath}/logout`}>Log out</DropdownItem>
     </Dropdown>
   ) : null;
 };
--- a/frontend/src/components/NavBar/UserInfo/__tests__/UserInfo.spec.tsx
+++ b/frontend/src/components/NavBar/UserInfo/__tests__/UserInfo.spec.tsx
@@ -51,6 +51,23 @@ describe('UserInfo', () => {
     expect(logout).toBeInTheDocument();
   });
 
+  it('should render logout link with correct context path', async () => {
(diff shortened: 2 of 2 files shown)


frontend/src/components/NavBar/UserInfo/UserInfo.tsx
```diff
@@ -19,7 +19,7 @@ const UserInfo = () => {
         </S.Wrapper>
       }
     >
-      <DropdownItem href="/logout">Log out</DropdownItem>
+      <DropdownItem href={`${window.basePath}/logout`}>Log out</DropdownItem>
```
Haarolean, line 22: Why do we have to do this here explicitly? This works fine in other places but not `/logout` for some reason. It doesn't look nice if we'd have to do this magic in other places in the future as well.

Haarolean (changes requested):

frontend/src/components/NavBar/UserInfo/UserInfo.tsx
pandeysambhi: let me take a look again

germanosin: @Haarolean This fix is correct. The `/logout` endpoint will always be called as `domain.com/logout`. If a prefix is configured, it should instead be called as `domain.com/{basePath}/logout`.

Haarolean: @germanosin Sure, my point is that none of the URLs within the frontend code require explicitly using window.basePath, but here we do for some reason. I'd like to know why :)

Haarolean: Aight, it seems it's the only react-app local url in the whole app used with href rather than other methods, thus it works differently requiring context path appended manually.

Haarolean (approved):

Haarolean:
@pandeysambhi thanks for your first PR! This will be merged once CI passes.

PR #1383, opened 2025-09-25
Change mock json to use alwaysFakeOptionals=false

HansMartinJ:
Why: For some complex schemas, the mock library hangs because of this option is true.

This is possibly because of a bug leading to an infinite loop, or because the nature of the schema leading to exponential amount of optional fields to be generated

Setting alwaysFakeOptionals to false fixes this issue for the schema that caused the error

Set alwaysFakeOptionals=false

I understand there is a tradeoff here, but it completely breaks some of our topic, so it would be great if it could be merged.

Also note that the schema mocker for some reason runs on the main topic page load (not just by clicking on produce)
So it stops the topic completely from being accessed.

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

Tested the schema locally with the schema mock library, tried with and without this option, where without the schema generated fine

🦭

### DIFF
--- a/frontend/src/components/Topics/Topic/SendMessage/utils.ts
+++ b/frontend/src/components/Topics/Topic/SendMessage/utils.ts
@@ -12,7 +12,7 @@ import AjvDraft4 from 'ajv-draft-04';
 import addFormats from 'ajv-formats';
 
 jsf.option('fillProperties', false);
-jsf.option('alwaysFakeOptionals', true);
+jsf.option('alwaysFakeOptionals', false);
 jsf.option('failOnInvalidFormat', false);
 
 const generateValueFromSchema = (preferred?: SerdeDescription) => {


Haarolean:
@HansMartinJ is that possible to cover this with a test?

HansMartinJ:
> @HansMartinJ is that possible to cover this with a test?

Not sure, I can't share the schema that caused it. We have a fork that we can fix it in so you may close this if needs more work, then you know the fix if it ever comes up again :)

Haarolean:
> > @HansMartinJ is that possible to cover this with a test?
> 
> Not sure, I can't share the schema that caused it. We have a fork that we can fix it in so you may close this if needs more work, then you know the fix if it ever comes up again :)

I don't mind merging this but a test would be nice so we don't get a regression later. Couldn't you anonimize fields in the schema?

HansMartinJ:
The schema is massive so its hard to fake it all

Anyways i think this is a bad solution, because we actually had to also had to add the following as well to fix the massive schema (like 9000000 fields were created even without fake optionals)
`jsf.option('maxItems', 5);`

 It may be better to expose the mock schema options to the user to give more control.

I'm closing it because it only a half-solution

PR #1412, opened 2025-10-08
FE: Hide internal topics by default (closes #1073)

### LINKED ISSUE #1073: Internal topics are not hidden by default when browsing topics in Kafka UI
### Issue submitter TODO list

_No response_

Describe the bug:
When using Kafka UI, internal topics (such as __consumer_offsets, etc.) are not hidden by default. Upon logging in and navigating to the “Browse Topics” section, all topics—including internal ones—are displayed automatically. This behavior occurs even if the internal topics were previously hidden or filtered out in a prior session.
⸻
Expected behavior:
Internal topics should ideally be hidden by default, or there should be an option in the configuration or UI settings to toggle this behavior. This would help improve usability and reduce clutter for users who don’t need to view internal topics regularly.
⸻
To Reproduce:
1. Log in to Kafka UI
2. Navigate to the “Browse Topics” section
3. Observe that internal topics are visible by default
4. Manually apply a filter to hide them (if desired)
⸻
Environment:
• Kafka UI Version: v1.2.0
⸻
Additional context:
This isn’t a major bug, but more of a usability concern. Especially in environments with many topics, seeing internal ones mixed in by default makes navigation a bit harder. Having an option in the UI or config file (e.g., hideInternalTopics: true) would be a great improvement.

_No response_

### Version you're running

v1.2.0

### Additional context

_No response_

Devansh-ops:
* [ ] **Breaking change?**
  **No.** Defaults only apply when the URL lacks `hideInternal` and no prior preference exists in `localStorage`. Existing explicit URL values and saved preferences continue to work as before.

* Frontend (Topics → `ListPage.tsx`): When the URL has no `hideInternal` param, derive the value from `localStorage('hideInternalTopics')`, **defaulting to `true` if missing**.
* Respect explicit URL values (`?hideInternal=true/false`).
* Parse the stored value as a real boolean (`'true'`/`'false'`) instead of truthy/falsy.
* Avoid mutating the existing `URLSearchParams` instance (create a new one before calling `setSearchParams`).

Closes #1073.

* URL precedence over localStorage (explicit URL should win).
* Whether we should **persist** the default (`'true'`) back to `localStorage` on first run (current change does, for consistency across sessions).
* Any preference on naming/placement if you’d like this logic factored into a small helper.

* [ ] No need to
* [x] Manually (IST):

  * Clear `localStorage`; open Topics → `?hideInternal=true` gets appended.
  * Toggle to show internal; reload persists as expected.
  * Visit with `?hideInternal=false` in URL; list shows internals; reload preserves via URL/localStorage.
* [ ] Unit checks
* [ ] Integration checks
* [ ] Covered by existing automation

* [ ] I have performed a self-review of my own code
* [ ] I have commented my code, particularly in hard-to-understand areas
* [ ] I have made corresponding changes to the documentation (e.g. **ENVIRONMENT VARIABLES**) — *note: UI behavior now defaults to hiding internal topics; added a short note to docs if required*
* [ ] My changes generate no new warnings (e.g. Sonar is happy)
* [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
* [ ] New and existing unit tests pass locally with my changes
* [ ] Any dependent changes have been merged

cute-bush-baby

### DIFF
--- a/frontend/src/components/Topics/List/ListPage.tsx
+++ b/frontend/src/components/Topics/List/ListPage.tsx
@@ -31,7 +31,7 @@ const ListPage: React.FC = () => {
       if (stored === null) localStorage.setItem('hideInternalTopics', 'true');
     } else {
       // sync localStorage if URL has it set
-      const raw =  searchParams.get('hideInternal');
+      const raw = searchParams.get('hideInternal');
       const norm = raw === 'true' || raw === 'false' ? raw : 'true'; // default to true if malformed
       localStorage.setItem('hideInternalTopics', norm);
       if (norm !== raw) searchParams.set('hideInternal', norm); // sync URL if malformed
--- a/frontend/src/components/Topics/List/ListPage.tsx
+++ b/frontend/src/components/Topics/List/ListPage.tsx
@@ -22,19 +22,27 @@ const ListPage: React.FC = () => {
     if (!searchParams.has('perPage')) {
       searchParams.set('perPage', String(PER_PAGE));
     }
-    if (
-      !!localStorage.getItem('hideInternalTopics') &&
-      !searchParams.has('hideInternal')
-    ) {
-      searchParams.set('hideInternal', 'true');
+    // If URL doesn't specify it, derive from localStorage (default = true when missing)
+    if (!searchParams.has('hideInternal')) {
+      const stored = localStorage.getItem('hideInternalTopics');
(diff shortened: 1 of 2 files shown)


Haarolean:
@Devansh-ops will this be ready for review any time soon? I'd like to close stale drafts

PR #1471, opened 2025-11-05
Dinf 2950 Fix schema prefix removal and ensure consistent topic name tranformation

ashok-pnt:
**Summary**
This PR fixes the schema subject mismatch issue by removing environment prefixes (like `development.` or `staging.`) from topic names before performing schema lookup.

**Key Changes**
- Updated `SchemaRegistrySerde.java` to strip environment prefix before schema lookup.
- Updated `TopicsExporter.java` to maintain consistent topic name transformation logic.
- Enhanced `SchemaRegistrySerdeTest.java` to validate schema retrieval, serialization/deserialization, and environment prefix removal.

**Validation**
Ran Gradle tests successfully — all unit tests passed, including `SchemaRegistrySerdeTest`.
Verified Kafka UI can now correctly deserialize schemas for environment-prefixed topics.
Confirmed that schema subjects match as expected after the fix.

**Impact**
Ensures consistent schema lookup across all environments and prevents `NOT FOUND` errors during deserialization.

### DIFF
--- a/api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
+++ b/api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
@@ -246,8 +246,9 @@ public class SchemaRegistrySerde implements BuiltInSerde {
     // Transform topic name from {env}.{topic_name} to {topic_name}
     // Example: "development.integration_events.connection_management.entities.connection" 
     // becomes "integration_events.connection_management.entities.connection"
-    String transformedTopic = topic.contains(".") ? 
-        topic.substring(topic.indexOf(".") + 1) : topic;
+    String transformedTopic = topic.contains(".")
+        ? topic.substring(topic.indexOf(".") + 1)
+        : topic;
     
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
```diff
@@ -243,7 +243,15 @@ private <T> Optional<T> wrapWith404Handler(Callable<T> call) {
   }
 
   private String schemaSubject(String topic, Target type) {
-    return String.format(type == Target.KEY ? keySchemaNameTemplate : valueSchemaNameTemplate, topic);
+    // Transform topic name from {env}.{topic_name} to {topic_name}
+    // Example: "development.integration_events.connection_management.entities.connection" 
+    // becomes "integration_events.connection_management.entities.connection"
+    String transformedTopic = topic.contains(".")
```
Haarolean, line 249: You have a wrong assumption that every user of the app has the same schema naming as you do. We won't proceed with such changes

Issue #1454, opened 2025-10-30
MCP server  doesn't properly handle Flux responses.


elvis-cai:
### Issue submitter TODO list

- [x] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [x] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md) and the feature is not present there

### Is your proposal related to a problem?

Endpoints that return Flux responses, including listAcls, listQuotas, listStreams, and others. 
Jackson can't serialize an active Flux stream, so it serializes the Flux's internal metadata properties instead, which is  {"scanAvailable": true, "prefetch": -1}.

<img width="607" height="244" alt="Image" src="https://github.com/user-attachments/assets/11dd35e7-6d70-46c9-b395-bc9c873cc6cd" />

### Describe the feature you're interested in

fix the mcp serialize issue

### Version you're running

ghcr.io/kafbat/kafka-ui:v1.3.0

PR #1491, opened 2025-11-12
BE: Handle Flux response bodies in MCP tool call results

elvis-cai:

  Problem:
  When MCP tools return Mono<ResponseEntity<Flux<?>>>, the response body contains a Flux stream that wasn't being properly handled, which
  could cause issues when trying to serialize the result for the MCP client.

  Solution:
  Modified the reponseToCallResult method to detect when the ResponseEntity body is a Flux and automatically collect it into a list before
   converting to CallToolResult. This ensures streaming responses are properly materialized before being serialized.

  Changes:
  - In McpSpecificationGenerator.java:93-101, added a check for Flux<?> response bodies
  - When detected, the Flux is collected via flux.collectList().map(this::callToolResult)
  - Non-Flux responses continue to work as before

  This allows MCP tools to return streaming responses (Flux) wrapped in ResponseEntity while ensuring they're properly handled and sent
  back to the MCP client as serialized lists.
fix  (https://github.com/kafbat/kafka-ui/issues/1454)

- [ ] No need to
- [] Manually (please, describe, if necessary)
- [] Unit checks
- [] Integration checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of successful API responses containing reactive data streams.
  * Reactive response content is now resolved correctly instead of being returned in an unprocessed format.
  * Empty reactive responses now produce a valid tool result instead of completing without a response.


api/src/main/java/io/kafbat/ui/service/mcp/McpSpecificationGenerator.java
```diff
@@ -93,7 +93,12 @@ private Mono<CallToolResult> toCallResult(Object result) {
   private Mono<CallToolResult> reponseToCallResult(ResponseEntity<?> response) {
     HttpStatusCode statusCode = response.getStatusCode();
     if (statusCode.is2xxSuccessful() || statusCode.is1xxInformational()) {
-      return Mono.just(this.callToolResult(response.getBody()));
+      Object body = response.getBody();
+      // Handle Flux response bodies by collecting them into a list
+      if (body instanceof Flux<?> flux) {
```
germanosin: I would prefer to see this trick in the callToolResult method

germanosin (dismissed):
Thanks for contribution. Please check my comment

api/src/main/java/io/kafbat/ui/service/mcp/McpSpecificationGenerator.java
elvis-cai: thanks for the quick review, good point, updated

Haarolean (approved):

kanchan78:
Hi @germanosin @Haarolean — since this is now approved, could someone with push access update the branch against \`main\`? It's currently reported as \`BEHIND\`, and the required \`Task Completed Checker\` check is failing from the stale Nov 2025 run — looks like a false positive from the "How Has This Been Tested?" section being a pick-one list (\`Covered by existing automation\` is checked, but the other mutually-exclusive options like \`No need to\` are correctly left unchecked, which the checklist bot misreads as incomplete). A rebase/merge from \`main\` should clear both and let it merge.

This issue has been open since October 2025, so it'd be great to get it landed. If for whatever reason this one can't be moved forward, #1952 has the identical fix plus a regression test covering the bug, and could be used instead — happy to have that closed either way, just flagging it's on standby if useful.

kanchan78:
Following up — this is still \`BLOCKED\` on the same \`Task Completed Checker\` failure as before (\`No need to\` unchecked). Looked closer at it: the "How Has This Been Tested?" section in the PR template is wrapped in \`<!-- ignore-task-list-start -->\` / \`<!-- ignore-task-list-end -->\` markers, same as the "Breaking change?" line above it — so it looks like it's meant to be excluded from the completeness check (it's a pick-one list, not an all-must-be-checked one). The checker doesn't seem to be respecting those markers here, so it's flagging a box that was correctly left unchecked.

Since it's already approved and otherwise mergeable, easiest unblock is probably either:
- @elvis-cai checking that box too (even though it's redundant next to "Covered by existing automation"), just to satisfy the linter, or
- a maintainer merging past the check / re-running it, given it's a known false positive.

Happy to be wrong here if there's more context I'm missing — just flagging since it's been sitting on this for a while.

PR #1542, opened 2025-11-29
Infra: Create more targeted groups for gradle and group all the action upgrades

yeikel:

Updated the Dependabot configuration to define more targeted groups. 

The intent is to increase the number of Dependabot pull requests that successfully merge into `main`, as the current grouping is too ambitious and frequently breaks the build (see #1520) causing us to simply ignore the updates and not get the main benefit of Dependabot

The action group was created because it’s generally safer to bundle these updates together. Depending on how things progress, we may need to apply a similar adjustment to it as well. 

I tested the groups in my fork: https://github.com/yeikel/kafka-ui/pulls

image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,7 +34,6 @@ updates:
         - "org.testcontainers:*"
         - "org.junit.jupiter:*"
         - "org.assertj:*"
-        - "com.gorylenko.gradle-git-properties"
         - "com.bmuschko.docker-remote-api"
         - "org.mockito:*"
         # Bouncy Castle is only used for testing purposes


yeikel:
@Haarolean Could you check this out?

.github/dependabot.yml
```diff
@@ -16,25 +16,55 @@ updates:
     # In general, our Netty references are temporary overrides, usually applied to address transitive Spring vulnerabilities, and should be configured with caution
     # In general, having conflicting Netty versions in the classpath is not recommended
     - dependency-name: "io.netty:*"
+    # We will handle major upgrades manually
```
Haarolean, line 19: i don't mind having PRs for these if they don't hog up the total amount of open PRs. Otherwise nobody will ever check if there's a new spring boot out

.github/dependabot.yml
```diff
@@ -16,25 +16,55 @@ updates:
...
         - "io.spring.dependency-management"
-      # We will handle major upgrades manually
-      update-types:
-        - "patch"
-        - "minor"
-    other-dependencies:
-      exclude-patterns:
-        - "org.springframework.boot:*"
-        - "io.spring.dependency-management"
+        - "io.modelcontextprotocol.sdk:mcp-spring-webflux"
```
Haarolean, line 28: why is this suddently a spring boot dependency?

.github/dependabot.yml
```diff
@@ -16,25 +16,55 @@ updates:
...
+        # We defined this dependency explicitly because Spring uses an older version
+        - "com.nimbusds:nimbus-jose-jwt"
+    testing:
+      # TODO: Consider using dependency-type when https://github.com/dependabot/dependabot-core/issues/13122 is supported
+      patterns:
+        - "org.testcontainers:*"
+        - "org.junit.jupiter:*"
+        - "org.assertj:*"
+        - "com.gorylenko.gradle-git-properties"
```
Haarolean: `"com.gorylenko.gradle-git-properties"` that package is for adding git meta info into the build, not testing

.github/dependabot.yml
yeikel: It is not, but I deducted from the name that it is probably a good idea to test these together as `mcp-spring-webflux` depends on Spring Webflux

.github/dependabot.yml
yeikel: The challenge with Dependabot is that you can enable either major or minor upgrades, but not both at the same time. If we enable major upgrades, only those will appear, and the smaller, more manageable minor upgrades will no longer be shown. 

In the case of Spring Boot, we will see Spring Boot 4.x immediately but no minor versions like 3.5.8. So while we wait to allocate capacity for the 4.x bump, we will need to manage minor versions manually

.github/dependabot.yml
yeikel: Thanks, I fixed this with a334ab7ccf5143f76b7d8d2f314817a25b9854a7

.github/dependabot.yml
yeikel: Do you think it should be separate?

.github/dependabot.yml
Haarolean: are you sure? can't we define multiple groups for that?

Haarolean: @copilot is that possible to enable dependabot PRs for spring so it creates both major and and minor semver PRs separately?

.github/dependabot.yml
Haarolean: ok makes sense!

.github/dependabot.yml
yeikel: Yes, I am pretty sure about this. See https://github.com/dependabot/dependabot-core/issues/13665

yeikel: What we will see separately and at the same time are security updates if any. But not all minor or patch releases will be security-related

Haarolean: Ok thank you! Do you think we should try other bots like renovate to circumvent this issue?

yeikel: I  am not sure if Renovate can do this, but we can check. 

I am active Dependabot contributor and I am _biased_, but I understand the need and the sentiment. Can we create and track a separate issue for this? I can take a look and let you know

Can we merge this in the meantime?

yeikel: I logged #1558

Haarolean: yes yes, sure, it's just a random thought: we don't have to do this nor or at all

PR #1568, opened 2025-12-10
Infra: use dependabot-friendly version references

yeikel:

Updates the GitHub Action version references to be Dependabot-friendly. That will allow Dependabot to update the documentation along with the version update in changes like https://github.com/kafbat/kafka-ui/pull/1566

- [x] No need to

I did not change any of the referenced versions. I only updated the documentation

image

### DIFF
--- a/.github/workflows/backend_tests.yml
+++ b/.github/workflows/backend_tests.yml
@@ -19,14 +19,14 @@ jobs:
     steps:
 
       - name: Checkout
-        uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2 # infered from @v4
+        uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2 # https://github.com/actions/checkout/releases/tag/v4.0.0
         with:
           token: ${{ github.token }}
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Set up JDK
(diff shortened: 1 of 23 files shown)


yeikel:
@Haarolean Could you please review this?

.github/workflows/backend_tests.yml
```diff
@@ -19,14 +19,14 @@ jobs:
     steps:
 
       - name: Checkout
-        uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2 # infered from @v4
+        uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2 # https://github.com/actions/checkout/releases/tag/v4.0.0
```
yeikel, line 22: Dependabot supports various formats. The other option is this 


```suggestion
        uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2 # v4.0.0
```

I personally prefer the version I committed because it is a link that resolves instead of a link we need to form manually

Haarolean: Not sure it does, it has failed here: https://github.com/kafbat/kafka-ui/pull/1527/files
so I'd prefer links

yeikel: That's because `# infered from @v4...` is not the right syntax. It would need to be something like `# v4`

PR #1594, opened 2025-12-23
Infra: Remove `io.modelcontextprotocol.sdk:mcp-spring-webflux` from the Spring boot group

yeikel:

In [https://github.com/kafbat/kafka-ui/pull/1542](https://github.com/kafbat/kafka-ui/pull/1542), we chose to include the `io.modelcontextprotocol.sdk:mcp-spring-webflux` dependency under the Spring Boot group, since it depends on Spring Boot and it is reasonable to test and upgrade them together.

One challenge with this dependency is that it introduces breaking changes in minor releases, which can unexpectedly break minor and patch-level Spring Boot dependency upgrades (for example, https://github.com/kafbat/kafka-ui/pull/1591
).

While it is permitted to create breaking changes under semantic versioning rules for 0.x versions, Dependabot does not differentiate this case and treats such updates as regular patch releases.

To avoid breaking our group and not getting the full benefit of Dependabot, we'll handle this dependency separately for now until it reaches 1.0.0

- [x] No need to

image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,7 +25,6 @@ updates:
       patterns:
         - "org.springframework.boot:*"
         - "io.spring.dependency-management"
-        - "io.modelcontextprotocol.sdk:mcp-spring-webflux"
         # We defined this dependency explicitly because Spring uses an older version
         - "com.nimbusds:nimbus-jose-jwt"
     testing:


yeikel:
@Haarolean Could you please check this out?

yeikel:
Thank you, @Haarolean 

Merging this unblocked https://github.com/kafbat/kafka-ui/pull/1596 🚀

PR #1599, opened 2025-12-28
CVE: Reduced failure level to high

germanosin:

- [x] No need to
- [ ] Unit checks
- [ ] Integration checks

### DIFF
--- a/.github/workflows/cve_checks.yml
+++ b/.github/workflows/cve_checks.yml
@@ -71,6 +71,7 @@ jobs:
         uses: aquasecurity/trivy-action@dc5a429b52fcf669ce959baa2c2dd26090d2a6c4 # https://github.com/aquasecurity/trivy-action/releases/tag/0.33.1
         with:
           image-ref: "ghcr.io/kafbat/kafka-ui:latest"
+          severity: "CRITICAL,HIGH"
           format: "table"
           exit-code: "1"
 
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -41,6 +41,7 @@ dependencies {
     implementation libs.spring.security.ldap
 
     implementation (libs.kafka.clients) {
+        // TODO: Remove once client would fix CVE
         exclude group: "org.lz4", module: "lz4-java"
     }
 
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -7,6 +7,16 @@ plugins {
     alias(libs.plugins.spring.dependency.management)
 }
 
+configurations.all {
+    resolutionStrategy {
+        capabilitiesResolution {
+            withCapability("org.lz4:lz4-java") {
+                select(libs.lz4.yawk.get().toString())
+            }
+        }
+    }
+}
+
 
 import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
 
@@ -74,10 +84,11 @@ dependencies {
     // CVE Fixes
     implementation libs.apache.commons.compress
     implementation libs.okhttp3.logging.intercepter
+    implementation libs.lz4.yawk
 
     implementation libs.modelcontextprotocol.spring.webflux
     implementation libs.victools.jsonschema.generator
-    implementation libs.lz4
+
 
     // Google Managed Service for Kafka IAM support
     implementation (libs.google.managed.kafka.login.handler) {
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -77,6 +77,7 @@ dependencies {
 
     implementation libs.modelcontextprotocol.spring.webflux
     implementation libs.victools.jsonschema.generator
+    implementation libs.lz4
 
     // Google Managed Service for Kafka IAM support
     implementation (libs.google.managed.kafka.login.handler) {
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -11,7 +11,7 @@ configurations.all {
     resolutionStrategy {
         capabilitiesResolution {
             withCapability("org.lz4:lz4-java") {
-                select(libs.lz4.get().toString())
+                select(libs.lz4.yawk.get().toString())
             }
         }
     }
@@ -86,7 +86,7 @@ dependencies {
     // CVE Fixes
     implementation libs.apache.commons.compress
     implementation libs.okhttp3.logging.intercepter
-    implementation libs.lz4
+    implementation libs.lz4.yawk
 
     implementation libs.modelcontextprotocol.spring.webflux
     implementation libs.victools.jsonschema.generator
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -11,7 +11,7 @@ configurations.all {
     resolutionStrategy {
         capabilitiesResolution {
             withCapability("org.lz4:lz4-java") {
-                select(libs.lz4.yawk.get().toString())
+                select(libs.lz4.get().toString())
             }
         }
     }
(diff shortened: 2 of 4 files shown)


.github/workflows/cve_checks.yml
```diff
@@ -71,6 +71,7 @@ jobs:
         uses: aquasecurity/trivy-action@dc5a429b52fcf669ce959baa2c2dd26090d2a6c4 # https://github.com/aquasecurity/trivy-action/releases/tag/0.33.1
         with:
           image-ref: "ghcr.io/kafbat/kafka-ui:latest"
+          severity: "CRITICAL,HIGH"
```
yeikel, line 74: What is the intention behind this change? The default severity set is `UNKNOWN, LOW, MEDIUM, HIGH, CRITICAL`. If the goal is to reduce noise, introducing separate or optional checks seems safer than narrowing the default. 

Given how inconsistent CVE scoring can be, I am concerned that hiding lower-severity findings risks causing more issues than it solves.

api/build.gradle
```diff
@@ -7,6 +7,16 @@ plugins {
     alias(libs.plugins.spring.dependency.management)
 }
 
+configurations.all {
+    resolutionStrategy {
+        capabilitiesResolution {
+            withCapability("org.lz4:lz4-java") {
```
yeikel, line 13: Can we add an inline comment/reference pointing to https://seclists.org/oss-sec/2025/q4/222 for this? Otherwise, it looks a bit cryptic

yeikel (changes requested):
Thanks for the change! 

I am not approving this because I feel strongly about the scanner behavior and would prefer to leave it as is, even if it results in additional noise, as filtering introduces the risk of missing relevant CVEs.

Severity classifications can vary between scanners. What is considered medium in one may be rated lower or higher in another, so applying filters could do more harm than good, in my opinion.

Aside from that, the change looks reasonable overall, with only a few minor documentation improvements to consider

api/build.gradle
```diff
@@ -30,7 +40,9 @@ dependencies {
 
     implementation libs.spring.security.ldap
 
-    implementation libs.kafka.clients
+    implementation (libs.kafka.clients) {
```
yeikel, line 43: This is tracked and fixed in https://issues.apache.org/jira/browse/KAFKA-19951, I assume that it may take a bit to land in Confluent's distribution. Can we document a TODO here to potentially remove this?  We may not even need the overwrite above after this is released depending on where else we use this library

.github/workflows/cve_checks.yml
germanosin: I I disagree with this approach. Medium- and low-level CVEs generate a lot of noise and force us to spend time fixing issues that provide little to no real value. This tool is designed for internal use, and I don’t think anyone is exposing it publicly. Closing all such issues makes sense for public-facing services, but for internal tools it’s largely a distraction. If someone does expose it publicly and needs these checks, we can maintain them in a separate repository.

yeikel: I think the main challenge is in how severity is defined. What one scanner considers *medium* and is currently filtered out may be considered *high* by another, specially if enterprise rules are in between to overwrite the CVE data. 

Perhaps a middle ground is to allow Medium issues as well?

> This tool is designed for internal use, and I don’t think anyone is exposing it publicly

Although this is probably true, I believe this is a risky assumption. Unless we clearly document it, we could be putting our users at risk assuming and designing around that. It’s difficult to predict how users might choose to deploy this or the ways in which they might do so

Beyond that, In enterprise environments, even when software is deployed entirely within an intranet network, scanners typically do not take deployment context into account as it just scans the deployed binaries. As a result, enterprise scanners will still flag these findings and expect users to address them, even when they may not have the ability to do so. This may lead to a higher number of security-related issues appearing in the repository. This has been the standard expectation in my organization, though I cannot speak for all environments, and we can evaluate and see how it goes

> force us to spend time fixing issues

I do agree on the noise point of view. But I do not understand this point. if the check is currently optional, it is just informational. In what way does it force us to do anything?

germanosin: Every failed pipeline forces the reviewer to investigate the root cause—whether the issue was introduced by the current code change or already existed—while also thinking through how to resolve it. At the same time, fixing CVEs is often a crunch-driven process, like with LZ4, where you end up pinning specific sub-dependency versions or switching to compatible forks. As a maintainer, spending this much time on each PR because of these issues is unacceptable. We should consider moving full CVE scanning out of the PR workflow and instead decide separately which low- and medium-severity issues should or need to be addressed.

yeikel: I think that we will have to agree to disagree here. I understand the challenge but I don't think that it is unacceptable spending time to ensure that the project is vulnerability-free and more so when the goal is to attract Enterprise users that often have no way to overwrite these. 

I suppose we can start from here and see how it evolves. Thanks!

yeikel (approved):

.github/workflows/cve_checks.yml
germanosin: I would still advocate for these changes. As a core maintainer, I genuinely believe they bring more value than drawbacks. I understand your concerns, and I’m not suggesting that security is unimportant. However, given the current stage of the product and how it is being used today, I would prioritize adding features at this point.

I’d appreciate hearing from other main contributors and backers—do you agree or disagree with this approach?

iliax: agree with @germanosin , for OSS product "CRITICAL,HIGH"  should be almost always enough. It also helps to make MRs easier for new contributors, which is more important on current project state (in my opinion).

Leshe4ka: I think that for an open-source project, prioritizing `CRITICAL` and `HIGH` vulnerabilities makes sense. Increasing the severity threshold better aligns with the current maturity of the product.

Haarolean: I agree with German:

- we’ve received multiple emails in the past with "fix these cves" messages. Not PRs, not sponsorships
- those emails usually boil down to "our scanner detected x/y/z", which adds little value. People run their own scanners anyway and are often unaware that we already have internal workflows in place.

I’d suggest leaving the reporting level to us and letting enterprise users engage in a more concrete way — via PRs or sponsorship.

germanosin: I appreciate the comments. I hope @yeikel will kindly consider the community’s opinion.

yeikel: Thank you all for the discussion. I understand the concern.

I may be one of the users who, in the future, reports CVEs that do not fall under the `CRITICAL` or `HIGH` categories due to the requirements of internal scanners. That said, you have been very accommodating in accepting my patches so far, and I have no concerns in that regard. My earlier points were more about the broader audience, who may not always be as understanding when interacting with you.

I also recognize that OSS development can be ungrateful and demanding at times, and that you have limited capacity to address every minor issue. With that in mind, I’m comfortable exploring this approach and seeing how it goes.

yeikel: I think that we should re-consider this and add `MEDIUM` back to the list 

CVE-2025-67735 is the first example we are missing as a result of this change. The impact to the project may be low(or not given it is Netty?)

yeikel: The good news is that for this specific example https://github.com/kafbat/kafka-ui/pull/1596 should be enough

Issue #240, opened 2024-03-22
Add option to select a time range


yeikel:
### Issue submitter TODO list

- [X] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [X] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/SECURITY.md) and the feature is not present there

### Is your proposal related to a problem?

Yes

### Describe the feature you're interested in

Currently, we have the option to select "Since time" and "To Time": 

![image](https://github.com/kafbat/kafka-ui/assets/8935151/c4f2b086-d75a-44d8-9739-14932562279b)

I would like to have the option to select both and build a time range that will show the messages that fall within that range.

For example : _Select messages between March 20 and March 22_

### Version you're running

2956664

YraganTron:
+1 to request

JacksonJang:
@Haarolean 
Could you please assign me to this issue?

JacksonJang:
@Haarolean @yeikel 
<img width="483" height="45" alt="Image" src="https://github.com/user-attachments/assets/16f381ee-cf04-4d28-ae45-6436d76dda94" />

My thought is to keep the current "Since time" approach as is,
and introduce an additional "time range" concept to support this use case.

What do you think about proceeding in this direction?

Haarolean:
> @Haarolean @yeikel 
> <img width="483" height="45" alt="Image" src="https://github.com/user-attachments/assets/16f381ee-cf04-4d28-ae45-6436d76dda94" />
> 
> My thought is to keep the current "Since time" approach as is,
> and introduce an additional "time range" concept to support this use case.
> 
> What do you think about proceeding in this direction?

sounds good, guess we could combine from and to filters altogether on the backend

PR #1608, opened 2026-01-04
FE: BE: Add option to select a time range

JacksonJang:
Issue: #240 

I added a new `TIMESTAMP_RANGE` polling mode to filter messages within a specific time range.

**Frontend:**
- Added `Time range` option to the polling mode selector
- Added dual date pickers for selecting start and end timestamps

**Backend:**
- Added `endTimestamp` query parameter to `/messages/v2` endpoint
- Added `TIMESTAMP_RANGE` to `PollingMode` enum
- Modified `SeekOperations`, `MessagesProcessing`, and `ForwardEmitter` to handle time range filtering

- The `endTimestamp` filtering logic in `MessagesProcessing.java` - messages are filtered after being read since Kafka doesn't support seek-to-end

- [ ] No need to
- [x] Unit checks
- [ ] Integration checks

image


JacksonJang:
@Haarolean @germanosin 
This has also passed all tests.

Haarolean:
@JacksonJang could you please take a look at these concerns raised by an LLM?

---

## PR #1608 Review: Add option to select a time range

**PR**: https://github.com/kafbat/kafka-ui/pull/1608
**Author**: JacksonJang | **Fixes**: #240 | **23 files** | **+257, -73** | **State**: Open

---

### Summary

Adds a new `TIMESTAMP_RANGE` polling mode that lets users filter messages within a start/end time range. The start timestamp determines where Kafka seeks to (like `FROM_TIMESTAMP`), and the end timestamp filters messages client-side in `MessagesProcessing`. The frontend adds dual date pickers when "Time range" is selected.

---

### Issues Found

#### 1. Pagination loses the end timestamp boundary (High)

**File**: `Cursor.java:62-68`

When creating a cursor for the next page, `TIMESTAMP_RANGE` maps to `FROM_OFFSET` with `endTimestamp` set to `null`:

```java
case FROM_OFFSET, FROM_TIMESTAMP, EARLIEST, TIMESTAMP_RANGE -> PollingModeDTO.FROM_OFFSET;
// ...
null, // endTimestamp lost here
```

This means page 2+ of a `TIMESTAMP_RANGE` query will return messages **beyond** the user's requested end time. Unlike `FROM_TIMESTAMP` (where losing the start boundary on page 2 is harmless since we've already passed it), losing the **end** boundary breaks the user's expected behavior.

**Suggestion**: Preserve `endTimestamp` in the cursor's `ConsumerPosition` and pass it through to the `ForwardEmitter`/`MessagesProcessing` on subsequent pages. This likely means `TIMESTAMP_RANGE` cursors should stay as `TIMESTAMP_RANGE` (not convert to `FROM_OFFSET`), or `endTimestamp` should be carried forward into the `FROM_OFFSET` position.

#### 2. `endTimestamp` filtering leaks to non-TIMESTAMP_RANGE modes (Medium)

**File**: `ConsumerPosition.java:61-78` and `ForwardEmitter.java:27`

`validateEndTimestamp()` returns `endTs` for **all** polling modes, not just `TIMESTAMP_RANGE`:

```java
private static Long validateEndTimestamp(PollingModeDTO pollingMode, ...) {
    if (pollingMode == PollingModeDTO.TIMESTAMP_RANGE) {
        // validation only for TIMESTAMP_RANGE
    }
    return endTs; // returned for ALL modes
}
```

And `ForwardEmitter` always passes `consumerPosition.endTimestamp()` to `MessagesProcessing`. Since `ForwardEmitter` handles `FROM_OFFSET`, `FROM_TIMESTAMP`, and `EARLIEST` too, a client passing `endTimestamp` with any of these modes would get silent end-timestamp filtering applied.

**Suggestion**: Return `null` for non-`TIMESTAMP_RANGE` modes:

```java
if (pollingMode != PollingModeDTO.TIMESTAMP_RANGE) {
    return null;
}
```

#### 4. No test for `TIMESTAMP_RANGE` in `SeekOperationsTest` (Medium)

**File**: `SeekOperationsTest.java`

Tests were updated to add the extra `null` parameter to `ConsumerPosition` constructors, but no test case was added for the actual `TIMESTAMP_RANGE` seek behavior. The `TIMESTAMP_RANGE` branch in `getOffsetsForSeek()` has two code paths (timestamp provided vs. null) — both should be covered.

#### 5. Weakened timestamp test assertions (Low)

**File**: `Filters.spec.tsx:201-214`

The existing tests for `TO_TIMESTAMP` and `FROM_TIMESTAMP` previously asserted the exact formatted date value. They now just assert the value is not empty:

```typescript
// Before:
expect(screen.getByPlaceholderText('Select timestamp')).toHaveValue(formattedDate);

// After:
expect(timestampInput).not.toHaveValue('');
```

This was presumably done because the date format changed from `"MMM d, yyyy"` to `"MMM d, yyyy HH:mm"`, making the old assertion invalid. But the new assertion is much weaker — it doesn't verify the date is correctly formatted or that the right date was rendered.

**Suggestion**: Update the expected format string rather than weakening the assertion.

#### 6. Date format change affects existing modes (Low — but good)

**File**: `Filters.tsx:608-616`

The `DatePickerInput` format changed from `"MMM d, yyyy"` to `"MMM d, yyyy HH:mm"` for `FROM_TIMESTAMP` and `TO_TIMESTAMP`. This is actually an improvement (more precision for timestamp-based filtering), but it's a behavioral change to existing features that should be noted in the PR description.

#### 7. `endTimestampReached` doesn't account for out-of-order timestamps (Low)

**File**: `MessagesProcessing.java:68-71`

```java
if (endTimestamp != null && rec.timestamp() > endTimestamp) {
    endTimestampReached = true;
    return;
}
```

Once `endTimestampReached` is set, all subsequent records in the current and future batches are skipped. This works correctly when records are globally sorted by timestamp (which `sortForSending` does via merge-sort). However, if a partition has out-of-order timestamps (possible with producer-set timestamps), a record with `ts > endTimestamp` could precede a record with `ts <= endTimestamp` within the same partition, causing valid records to be dropped.

This is a minor edge case that also affects other timestamp-based modes and is acceptable for a first implementation.

#### 8. No integration test for end-to-end TIMESTAMP_RANGE flow (Low)

There's no `RecordEmitterTest` or `MessagesServiceTest` case that verifies the complete `TIMESTAMP_RANGE` flow — producing messages with known timestamps, querying with a range, and asserting only in-range messages are returned. This would catch issues like #1 (pagination) and provide confidence in the feature.

---

### What Looks Good

- **Clean backend architecture**: Reusing `ForwardEmitter` with `FROM_TIMESTAMP` seek logic for the start boundary and adding client-side filtering for the end boundary is pragmatic — Kafka doesn't support seek-to-end-timestamp natively
- **Frontend UX**: Dual date pickers with `selectsStart`/`selectsEnd` and `minDate` constraints are well implemented, and the `~` separator between pickers is intuitive
- **URL param persistence**: `endTimestamp` is correctly persisted/restored from URL params and local storage, consistent with existing filter handling
- **Mode selector cleanup**: Extracting `renderModeInput()` into a dedicated function improves readability over the previous inline conditionals
- **Test updates**: All existing tests were updated to accommodate the new `ConsumerPosition` field

---

### Verdict

The core feature is well-designed and the approach is sound. However, **issue #1** (pagination losing end timestamp) is a functional bug that will affect users as soon as they paginate, and **issue #3** (missing OpenAPI YAML update) will prevent the feature from building correctly. These should be addressed before merge. Issue #2 (endTimestamp leaking to other modes) is a straightforward fix that should also be included.

Haarolean (changes requested):
please check my previuos comment

Haarolean:
please ping me if you wish to recover the branch

PR #1685, opened 2026-02-10
fix(messages): prevent infinite re-read loop when polling range is empty

### LINKED ISSUE #1684: Huge consumed bytes/messages
### Issue submitter TODO list

**Description**

When loading topic messages with a cursor (e.g. "Next page"), if the cursor has `limit = 0` (or `messagesPerPage` is 0 for any other reason), the polling logic builds an empty range `(from, from)`. The next range is also `(from, from)`, so the consumer never advances and keeps re-reading the same data until the request is cancelled or times out.

**Observed impact:**

- One request ran for **~73 minutes** (4,382,435 ms), read **202 GB** and **433,048,747 messages consumed**, then **completed** (e.g. by client/server timeout or connection close).
- The UI showed **"No messages found"** — **zero messages** were returned to the user.
- The topic has on the order of **~1M messages**; 433M consumed means the same data was effectively read hundreds of times in a single request.

So the bug causes both unbounded load (hundreds of GB, hundreds of millions of records) and a useless result (no messages).

---

**Actual behavior**

- When `messagesPerPage == 0`, `msgsToPollPerPartition` becomes 0.
- The range is built as `(fromOffset, fromOffset + 0)` → `(from, from)` (empty range).
- In the inner loop, only records with `offset < fromTo.to` are added to the result; with `to == from`, **no records** satisfy this, so **result is always empty**.
- The client receives **zero MESSAGE events** → UI shows **"No messages found"**.
- Every iteration still calls `poll()`; all those records are counted in "messages consumed" and bytes.
- `nextPollingRange` returns the same `(from, from)` again, so the loop never advances and keeps re-reading the same data until the request is cancelled or times out (e.g. after tens of minutes).
- After the request ends, the UI shows very large consumption stats (e.g. 433M messages, 202 GB) and an empty message list.

---

**Root cause**

1. **ForwardEmitter / BackwardEmitter**  
   `msgsToPollPerPartition` is computed as  
   `(int) Math.ceil((double) messagesPerPage / readFromOffsets.size())`.  
   When `messagesPerPage == 0`, this is **0**, so the range becomes `(fromOffset, fromOffset)`. The consumer seeks to `from`, reads batches, but no record has `offset < from`, so the result list stays empty and no messages are sent to the client. The next range is again `(from, from)`, so the loop never moves forward.

2. **MessagesService**  
   When loading by cursor, `cursor.limit()` is passed directly into `loadMessages` without normalization. So a cursor stored with `limit = 0` (or any path that passes 0) leads to `messagesPerPage = 0` in the emitter and triggers the empty range and the re-read loop.

---

**Suggested fix**

1. **Guarantee non-zero step in emitters**  
   In `ForwardEmitter` and `BackwardEmitter`, ensure the per-partition step is at least 1, e.g.  
   `int msgsToPollPerPartition = Math.max(1, (int) Math.ceil((double) messagesPerPage / readFromOffsets.size()));`  
   (and the same for `readToOffsets.size()` in BackwardEmitter). This prevents the range from ever being `(from, from)`.

2. **Normalize limit when loading by cursor**  
   In `MessagesService.loadMessages(KafkaCluster, String, String cursorId)`, pass a normalized limit (e.g. via the same logic as `fixPageSize(cursor.limit())`) so that 0 is never passed to the emitter.

3. **Optional safeguard**  
   In the polling loop, exit if total messages read exceeds the topic size (e.g. `seekOperations.summaryOffsetsRange() * (1 + margin)`) to avoid unbounded re-reading even if another bug produces a similar pattern.

---

**Affected code**

- `api/src/main/java/io/kafbat/ui/emitter/ForwardEmitter.java` (lines ~47, 50–52)
- `api/src/main/java/io/kafbat/ui/emitter/BackwardEmitter.java` (analogous)
- `api/src/main/java/io/kafbat/ui/service/MessagesService.java` (loadMessages with cursorId, ~235–245)

---

**Environment**

- Reproduced with a topic on the order of 1M messages, multiple brokers; request completed after ~73 minutes with 433M messages consumed, 202 GB read, and "No messages found" in the UI.

### Expected behavior

- Polling advances by a non-empty range each time and stops when the end of the topic is reached or the page limit is satisfied.
- A single request does not re-read the same offsets hundreds of times.
- If there are messages in the topic (and they match the filter), at least some messages are returned; if none match, the request still finishes after at most one pass over the topic.

### Your installation details

8b5494b
24.07.2025, 14:40:05

### Steps to reproduce

1. Open topic messages in Kafka UI and run a search with a filter.
2. Trigger loading the next page via cursor (e.g. "Next page" or any flow that uses the stored cursor).
3. If the cursor was stored with `limit = 0` (or 0 is passed as page size somewhere), the backend enters the infinite re-read loop.
4. Let the request run until it ends (timeout or disconnect). Observe: very high "messages consumed" and "bytes consumed", and **"No messages found"** in the message list.

### Screenshots

_No response_

### Logs

_No response_

### Additional context

_No response_

Mohoki:

When messagesPerPage (limit) was 0, msgsToPollPerPartition became 0, producing an empty range (from, from). nextPollingRange then returned the same range, so the consumer never advanced and re-read the same data until timeout, causing hundreds of millions of messages consumed and zero results (No messages found).

Changes:
- ForwardEmitter/BackwardEmitter: ensure msgsToPollPerPartition is at least 1 via Math.max(1, ...) so the range is never empty.
- MessagesService: when loading by cursor, pass fixPageSize(cursor.limit()) so a cursor with limit 0 is normalized to defaultPageSize and never reaches the emitter.

Tests:
- RecordEmitterTest: add forwardAndBackwardEmitterWithZeroMessagesPerPageCompleteWithoutHanging
- CursorTest: add forwardEmitterWithZeroLimitCompletesWithoutHanging and backwardEmitterWithZeroLimitCompletesWithoutHanging

- [ ] No need to
- [x] Unit checks
- [ ] Integration checks

Resolves: #1684

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ target/
 .settings
 .springBeans
 .sts4-cache
+.pnpm-store
 
 ### IntelliJ IDEA ###
 .idea
--- a/api/src/main/java/io/kafbat/ui/config/ClustersProperties.java
+++ b/api/src/main/java/io/kafbat/ui/config/ClustersProperties.java
@@ -110,6 +110,7 @@ public class ClustersProperties {
     Integer maxPageSize;
     Integer defaultPageSize;
     Integer responseTimeoutMs;
+    Integer maxMessagesToScanPerPoll;
   }
 
   @Data
--- a/api/src/main/java/io/kafbat/ui/emitter/AbstractEmitter.java
+++ b/api/src/main/java/io/kafbat/ui/emitter/AbstractEmitter.java
@@ -16,10 +16,21 @@ abstract class AbstractEmitter implements java.util.function.Consumer<FluxSink<T
     this.pollingSettings = pollingSettings;
   }
 
+  protected PollingSettings getPollingSettings() {
+    return pollingSettings;
+  }
+
   protected PolledRecords poll(FluxSink<TopicMessageEventDTO> sink, EnhancedConsumer consumer) {
     var records = consumer.pollEnhanced(pollingSettings.getPollTimeout());
     sendConsuming(sink, records);
     return records;
+  }
+
+  /**
+   * Poll without sending CONSUMING stats. Used by range emitters to report in-range-only stats.
+   */
+  protected PolledRecords pollWithoutConsuming(EnhancedConsumer consumer) {
+    return consumer.pollEnhanced(pollingSettings.getPollTimeout());
   }
 
   protected boolean isSendLimitReached() {
@@ -38,6 +49,10 @@ abstract class AbstractEmitter implements java.util.function.Consumer<FluxSink<T
 
   protected void sendConsuming(FluxSink<TopicMessageEventDTO> sink, PolledRecords records) {
     messagesProcessing.sentConsumingInfo(sink, records);
+  }
+
+  protected void sendConsumingInRange(FluxSink<TopicMessageEventDTO> sink, int inRangeRecords, long inRangeBytes, long elapsedMs) {
+    messagesProcessing.sentConsumingInfo(sink, inRangeRecords, inRangeBytes, elapsedMs);
   }
 
   // cursor is null if target partitions were fully polled (no, need to do paging)
--- a/api/src/main/java/io/kafbat/ui/emitter/BackwardEmitter.java
+++ b/api/src/main/java/io/kafbat/ui/emitter/BackwardEmitter.java
@@ -44,8 +44,11 @@ public class BackwardEmitter extends RangePollingEmitter {
       );
     }
 
-    int msgsToPollPerPartition = Math.max(1, (int) Math.ceil((double) messagesPerPage / readToOffsets.size()));
     TreeMap<TopicPartition, FromToOffset> result = new TreeMap<>(Comparator.comparingInt(TopicPartition::partition));
+    if (readToOffsets.isEmpty()) {
+      return result;
+    }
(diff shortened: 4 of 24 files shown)


Haarolean:
> When messagesPerPage (limit) was 0

please elaborate on how can this happen

Mohoki:
Good point.

`limit=0` shouldn’t happen in the normal first-page flow (there we already run `fixPageSize(...)`), but in the cursor flow we previously used `cursor.limit()` as-is. So if an old/invalid cursor had `0`, it went straight to emitters and triggered the zero-step behavior.

That’s why we fixed both:
- normalize limit in cursor path too;
- force emitter step to be at least 1.

At the same time, from newer logs we see another issue: offsets were moving and polling finished, so that case was not the `limit=0` loop. There’s a separate range-polling efficiency problem causing huge consumed data.

So this PR fixes a real edge case, but it’s not the full root-cause fix for the heavy/slow reads. Better not merge as “final fix” yet.

Haarolean:
can you provide steps to reproduce this issue? let's start with that.

Mohoki:
Hi @Haarolean, repro steps for the original issue:

Create a topic with many partitions and a large backlog (e.g. ~30 partitions, ~200k+ records total)
Open topic messages in UI
Apply a filter that not matches any record (or just one so UI keeps searching)
Trigger search and watch backend logs/consuming stats (messagesConsumed, bytesConsumed)

PR #1702, opened 2026-02-23
fix: use getAllVersionsById instead of getAllSubjectsById for Karapace compatibility

### LINKED ISSUE #1701: Avro deserialization broken with Karapace/non-Confluent Schema Registries after #1668
### Issue submitter TODO list

Avro message deserialization fails for **all** messages when using a Schema Registry implementation that does not support the `GET /schemas/ids/{id}/subjects` endpoint (e.g., Aiven Karapace).

Commit `181b423e` ("Expose subject names in the messages #1668") added a new call to `schemaRegistryClient.getAllSubjectsById(id)` in the `SchemaRegistrySerde.deserializer()` method. This call hits `GET /schemas/ids/{id}/subjects` on the Schema Registry. Karapace does not implement this endpoint and returns a 404, which propagates as a `RuntimeException` and kills deserialization for every Avro-encoded message.

The exception is thrown **before** `formatter.format()` is called, so no messages can be deserialized at all — not just those with missing subjects.

Relevant code in `SchemaRegistrySerde.java`:

```java
private List<String> getSubjectsById(int schemaId) {
    return idToSubjectsCache.get(schemaId, (id) -> {
      try {
        return schemaRegistryClient.getAllSubjectsById(id).stream() // <-- 404 on Karapace
            .filter(s -> !s.isEmpty())
            .toList();
      } catch (Exception e) {
        throw new RuntimeException(e); // <-- kills all deserialization
      }
    });
}
```

### Expected behavior

Avro deserialization should work with any Schema Registry implementation that supports the core Confluent SR API (schema lookup by ID). The subject name lookup should either:

1. Use `getAllVersionsById(id)` instead, which hits `GET /schemas/ids/{id}/versions` — an endpoint supported by **both** Confluent SR and Karapace — and extract subjects via `SubjectVersion::getSubject`
2. Or at minimum, catch the exception gracefully and fall back to an empty subject list, since subjects are metadata for display only and not required for deserialization

### Your installation details

- **App version**: `c7e60a39` (latest `main` as of this report)
- **Schema Registry**: Aiven Karapace (also affects any SR that doesn't implement `GET /schemas/ids/{id}/subjects`)
- **Regression introduced in**: commit `181b423e` (PR #1668)
- **Last known working commit**: `fc34cfe`

### Steps to reproduce

1. Configure kafbat-ui with a Karapace-based Schema Registry (e.g., Aiven managed Kafka)
2. Produce Avro-encoded messages to a topic (messages registered with SR as normal)
3. Open the topic in kafbat-ui and attempt to view messages
4. All Avro messages fail to deserialize

You can verify the root cause independently:
```bash
# This returns 404 on Karapace:
curl http://<schema-registry>:8081/schemas/ids/1/subjects

# This works on both Karapace and Confluent SR:
curl http://<schema-registry>:8081/schemas/ids/1/versions
```

### Screenshots

_No response_

## Logs

Expected stack trace in kafbat-ui logs:
```
RuntimeException wrapping RestClientException or IOException
  at SchemaRegistrySerde.getSubjectsById()
  at SchemaRegistrySerde$lambda.deserializer()
```

### Proposed fix

The Confluent `SchemaRegistryClient` interface provides both methods:
- `getAllSubjectsById(int)` → `Collection<String>` → hits `/schemas/ids/{id}/subjects` (Confluent-only)
- `getAllVersionsById(int)` → `Collection<SubjectVersion>` → hits `/schemas/ids/{id}/versions` (both Confluent and Karapace)

The fix is a one-line change in `SchemaRegistrySerde.getSubjectsById()`:

```
// Before:
return schemaRegistryClient.getAllSubjectsById(id).stream()
    .filter(s -> !s.isEmpty())
    .toList();

// After:
return schemaRegistryClient.getAllVersionsById(id).stream()
    .map(SubjectVersion::getSubject)
    .filter(s -> !s.isEmpty())
    .distinct()
    .toList();
```

This returns the same data, uses the same caching, and works on both Confluent SR and Karapace.

### Karapace API coverage

Verified against Karapace source ([`src/karapace/api/routers/schemas.py`](https://github.com/Aiven-Open/karapace/blob/main/src/karapace/api/routers/schemas.py)):

| Endpoint | Confluent SR | Karapace |
|---|---|---|
| `GET /schemas/ids/{id}` | Yes | Yes |
| `GET /schemas/ids/{id}/versions` | Yes | Yes |
| `GET /schemas/ids/{id}/subjects` | Yes | **No** |

### Related issues

- #1449 — Reports other Aiven/Karapace compatibility issues (403 on `/config`, connection errors). Different endpoints but same root theme: kafbat-ui assumes Confluent SR API coverage that Karapace doesn't fully provide.

### Secondary issue in same commit

The commit also renamed the deserialization metadata property from `"schemaId"` to `"id"`. If any consumers depend on the old key name, they will silently get `null`.

joshuaNathaniel:

Fixes #1701 

  Replaced `schemaRegistryClient.getAllSubjectsById(id)` with `schemaRegistryClient.getAllVersionsById(id)` in `SchemaRegistrySerde.getSubjectsById()`.

  `getAllSubjectsById` hits `GET /schemas/ids/{id}/subjects`, which is not implemented by Karapace (Aiven's open-source Schema Registry). This causes a 404 that propagates as a `RuntimeException`, breaking Avro deserialization for **all**
  messages.

  `getAllVersionsById` hits `GET /schemas/ids/{id}/versions`, which is supported by both Confluent SR and Karapace. The subject names are extracted via `SubjectVersion::getSubject` -  same data, same caching, just a compatible endpoint.

  The change is in `api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java`. It's a one-line functional change plus an import. Confirm that `getAllVersionsById` returns the expected subjects in your environment.

    - Stood up karapace locally to replicate issue. Solution works with both confluent and karapace
  - [x] Unit checks
  

  Manually verified with a Docker Compose environment running Karapace as the Schema Registry. Before the fix, all Avro messages failed to deserialize. After the fix, deserialization works correctly. `SchemaRegistrySerdeTest` and
  `SendAndReadTests` both pass.

=^._.^= ∫

### DIFF
--- a/api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
+++ b/api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
@@ -18,6 +18,7 @@ import io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient;
 import io.confluent.kafka.schemaregistry.client.SchemaMetadata;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClientConfig;
+import io.confluent.kafka.schemaregistry.client.rest.entities.SubjectVersion;
 import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException;
 import io.confluent.kafka.schemaregistry.json.JsonSchema;
 import io.confluent.kafka.schemaregistry.json.JsonSchemaProvider;
@@ -355,8 +356,10 @@ public class SchemaRegistrySerde implements BuiltInSerde {
   private List<String> getSubjectsById(int schemaId) {
     return idToSubjectsCache.get(schemaId, (id) -> {
       try {
-        return schemaRegistryClient.getAllSubjectsById(id).stream()
+        return schemaRegistryClient.getAllVersionsById(id).stream()
+            .map(SubjectVersion::getSubject)
             .filter(s -> !s.isEmpty())
+            .distinct()
             .toList();
       } catch (Exception e) {
         throw new RuntimeException(e);


Haarolean (approved):

Haarolean:
@joshuaNathaniel thank you

PR #1741, opened 2026-03-24
BE: support MessagePackSerde

### LINKED ISSUE #951: Serde: Support MessagePack
### Issue submitter TODO list

_No response_

I would be grateful to have MessagePack support and see it as one of the Serde dropdown selections 😄 
https://msgpack.org/index.html

_No response_

### Version you're running

4cf17a0 v1.1.0

### Additional context

_No response_

YeeaaahMan:
Fixes #951 
Add MessagePackSerde which supports msgpack format. https://msgpack.org/

Serialization and de serialization works, new option is available in Messages preview tab.

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

I used https://msgpack.solder.party/ to define basic objects: string, list, dict.
Then copied HEX representation to the Produce Message dialog.
As result all messages are displayed correctly when MessagePack serde is selected.
Base64-repr MessagePack-repr

## Summary by CodeRabbit

* **New Features**
  * MessagePack support: view, deserialize, and serialize MessagePack-encoded messages as human-readable JSON strings in the UI.

* **Tests**
  * Added tests covering MessagePack serialization/deserialization across target types, including success cases and error handling for invalid or empty inputs.


Haarolean:
@YeeaaahMan hi, thanks for your contirbution. As this serde is not quite a popular unlike schem registry or glue and it requires external dependencies, we'd rather see this as a separate standalone serde, as I suggested [here](https://github.com/kafbat/kafka-ui/issues/951#issuecomment-2745246204). 
Would you like to create your own repo and we list the serde as a community one or perhaps I could create a repo in our org so we could share maintenance overhead in the future?

YeeaaahMan:
I see. It sounds logical; this should be moved to a separate repository.

I’d prefer to leave this to you, as I’m completely unfamiliar with the Java ecosystem and can’t quite estimate the scope of work right now. Alternatively, we could hop on a call to outline a plan, and then I can try to handle it myself.

In any case, I’d be happy to help at least partially.

api/build.gradle
```diff
@@ -58,6 +58,7 @@ dependencies {
     implementation libs.confluent.avro.serializer
     implementation libs.confluent.protobuf.serializer
     implementation libs.confluent.json.schema.serializer
+    implementation 'org.msgpack:msgpack-core:0.9.11'
```
Haarolean: please extract the version into the catalog

Haarolean:
> I see. It sounds logical; this should be moved to a separate repository.
> 
> I’d prefer to leave this to you, as I’m completely unfamiliar with the Java ecosystem and can’t quite estimate the scope of work right now. Alternatively, we could hop on a call to outline a plan, and then I can try to handle it myself.
> 
> In any case, I’d be happy to help at least partially.


discussed with @germanosin — messagepack is quite popular and isn't dependency-heavy, we'll include it as a built-in serde. Please take a look at the inline comment I left, the rest is good to go.

Haarolean:
@YeeaaahMan also please add some tests

YeeaaahMan:
I found some inconsistencies in the serializer's behavior and removed it. 
Deserialization works correctly, and unit tests have been added.

If we decide to add a serializer (used in the "Produce Message" dialog), the logic should be:
```
String -> JSON object -> msgpack.to_bytes(data) -> bytes
```
This would require a new dependency: org.msgpack.jackson-dataformat-msgpack. But is it worth it? 

My main goal was to provide a convenient way to view messages in msgpack format, and the current deserializer handles that perfectly. Any other functionality can be added later.

Guys, what do you think?

Haarolean:
> I found some inconsistencies in the serializer's behavior and removed it. Deserialization works correctly, and unit tests have been added.
> 
> If we decide to add a serializer (used in the "Produce Message" dialog), the logic should be:
> 
> ```
> String -> JSON object -> msgpack.to_bytes(data) -> bytes
> ```
> 
> This would require a new dependency: org.msgpack.jackson-dataformat-msgpack. But is it worth it?
> 
> My main goal was to provide a convenient way to view messages in msgpack format, and the current deserializer handles that perfectly. Any other functionality can be added later.
> 
> Guys, what do you think?

I don't mind either personally. We could start with deserialization and implement serialization in a separate task later if requested

YeeaaahMan:
Serialization was added. Many corner cases were tested and covered by unit tests.
Checked on UI: it's impossible to add invalid data in various simple scenarios, output is ok. 
https://github.com/user-attachments/assets/0ee65cd0-d525-4b7c-8f7b-8651284dddb7

De-serialization also is ok.

Ready to merge.

Haarolean:
@coderabbitai review

api/src/main/java/io/kafbat/ui/serdes/builtin/MessagePackSerde.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public static final String FAILED_TO_DESERIALIZE_MSGPACK_PAYLOAD = "Failed to deserialize MessagePack payload";
+  public static final String FAILED_TO_SERIALIZE_JSON_PAYLOAD = "Failed to parse JSON payload";
+
+  private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
+  private static final ObjectMapper MSGPACK_MAPPER = new ObjectMapper(new MessagePackFactory());
+
+  @Override
+  public boolean canDeserialize(String topic, Serde.Target type) {
+    return true;
```
Haarolean, line 27: why generic true for both methods? can't we scan the message first bytes to detect if it's applicable?

YeeaaahMan: MessagePack doesn't have any common "magic byte" in the beginning of  sequence. Here is the specification:
https://github.com/msgpack/msgpack/blob/master/spec.md#formats 

It has various markers in bytes sequence for every type of object, but not for all (e.g. 7bit int, null, bool).

Haarolean: what about the ones in `EXPECTED_MESSAGE_PACK_BYTES`?

YeeaaahMan: The first 0x81 byte  is a marker of Map object (with 1 key)
https://github.com/msgpack/msgpack/blob/master/spec.md#map-format-family

For String of Array first byte will be another.

YeeaaahMan: You can play with online converter, to figure out how if works: https://msgpack.solder.party/
It's quite simple.

YeeaaahMan:
```java
public class MessagePackSerdeTest {
  private static final String EXPECTED_JSON_STRING = "{\"items\":[null,true,2]}";
  private static final byte[] EXPECTED_MESSAGE_PACK_BYTES = {
      (byte) 0x81, // 0x80 = fixmap + 0x01 = with 1 entry
      (byte) 0xa5, // 0xa0 = string + 0x05 = length 5
      (byte) 0x69, // 'i'
      (byte) 0x74, // 't'
      (byte) 0x65, // 'e'
      (byte) 0x6d, // 'm'
      (byte) 0x73, // 's'
      (byte) 0x93, // 0x90 = fixarray + 0x03 = array with 3 elements
      (byte) 0xc0, // null
      (byte) 0xc3, // true
      (byte) 0x02  // 2
  };
```

PR #1764, opened 2026-03-28
BE: MCP: Enforce cluster readOnly restriction for write operations

### LINKED ISSUE #1751: MCP bypasses cluster readOnly restriction
### Issue submitter TODO list

The `readOnly` cluster setting (`KAFKA_CLUSTERS_0_READONLY=true`) can be fully bypassed via MCP, if the MCP server is enabled, while the restriction is correctly enforced in the frontend. This has been confirmed for e.g. producing and deleting messages and topics.

Given the nature of the bypass (the WebFilter is never passed for any MCP tool call), it is reasonable to assume that every write operation exposed by the kafka-ui - modifying broker configs, managing ACLs, updating schemas, managing connectors - is equally unprotected.

### Expected behavior

The  `readOnly` cluster setting should be respected not only in the frontend / UI (API calls to `/api/clusters/...` endpoints), but also for any MCP interaction (e.g. `/mcp/message` or `/mcp/sse`). MCP clients should receive a meaningful error when writing tools fail.

### Your installation details

1. App version:  [dfa5a7e](https://github.com/kafbat/kafka-ui/commit/dfa5a7e)
2. Helm chart version: 1.6.0
3. Relevant config properties:
     ```
     KAFKA_CLUSTERS_0_READONLY: "true"
     MCP_ENABLED: "true"
     ```

### Steps to reproduce

1. Spin up a kafka-ui deployment in read-only mode and with MCP server enabled (see config above).
2. Connect MCP client and ask to run a destructive or creative action like producing or deleting messages.
3. Check if the action has been successfully performed, bypassing the read-only configuration.

### Screenshots

_No response_

### Logs

_No response_

### Root Cause

The readonly enforcement is implemented as a Spring `WebFilter`:

https://github.com/kafbat/kafka-ui/blob/dfa5a7e9068ea47b34bf7323dc7610432a4b1ab6/api/src/main/java/io/kafbat/ui/config/ReadOnlyModeFilter.java#L23-L71

The filter intercepts all non-GET HTTP requests, extracts the cluster name from the URL path using the regex `/api/clusters/(?<clusterName>[^/]++)` (lines 24–25), and throws a `ReadOnlyModeException` if the cluster is read-only. This works correctly for direct REST API calls to `/api/clusters/...`.

The MCP server does **NOT** route tool calls through the HTTP request pipeline. Tool dispatch is handled by `McpSpecificationGenerator`, which calls controller methods directly:

https://github.com/kafbat/kafka-ui/blob/dfa5a7e9068ea47b34bf7323dc7610432a4b1ab6/api/src/main/java/io/kafbat/ui/service/mcp/McpSpecificationGenerator.java#L64-L82

The `serverWebExchange` in context is the SSE/MCP connection exchange (registered at `/mcp/message` and `/mcp/sse`), not a new exchange for an `/api/clusters/...` endpoint:

https://github.com/kafbat/kafka-ui/blob/dfa5a7e9068ea47b34bf7323dc7610432a4b1ab6/api/src/main/java/io/kafbat/ui/config/McpConfig.java#L27-L31

Since `ReadOnlyModeFilter` only intercepts requests whose path matches `/api/clusters/`, the MCP dispatch path **never triggers it for any tool call**.

None of the individual controllers or services check `cluster.isReadOnly()` independently, they rely entirely on the filter. This means the bypass is not limited to a specific endpoint: **every write operation exposed as an MCP tool inherits the same vulnerability**.

### Impact

Any client with access to the MCP endpoint (`/mcp/sse`) can bypass the `readOnly` restriction entirely and perform any write operation available in kafka-ui.

The bypass is silent: all calls return success with no error or indication that a security control was circumvented. An administrator relying on `KAFKA_CLUSTERS_0_READONLY=true` to protect a cluster has no indication that the MCP server renders this control ineffective.

### Suggested Fix

Either of the following would address the issue:

1. **Enforce in MCP dispatch layer:** Add a readonly check in `McpSpecificationGenerator.methodCall()` before invoking the controller method, resolving the cluster name from the tool arguments and checking `clustersStorage.getClusterByName(...).isReadOnly()`.

2. **Enforce in controllers:** Have each write controller explicitly check `getCluster(clusterName).isReadOnly()` and return an error, rather than relying solely on the WebFilter. This makes the guard robust regardless of invocation path.

### Disclaimer

This issue has been created only after notifying maintainers via email first and clarifying that this design flaw does not classify as "security vulnerability" in the sense of the project's security policy and may be reported publicly.

--- 

Hope this write-up is useful, and thanks for maintaining such a great tool!

I'm not much of a java developer (or developer at all 😅), but I'm happy to test any proposed fix if that would be helpful. ✌️

paikend:
## Summary

Fixes #1751

MCP tool calls bypass the `ReadOnlyModeFilter` because they arrive via `POST /mcp/message` instead of `/api/clusters/{clusterName}/...` paths. This means write operations (create topic, delete schema, etc.) can be executed through MCP even when the cluster is configured as read-only.

### Root Cause

`ReadOnlyModeFilter` is a `WebFilter` that checks the HTTP request path against `/api/clusters/{clusterName}` pattern. MCP requests use a different path (`/mcp/message`, `/mcp/sse`), so they bypass this filter entirely.

### Solution

Added a readOnly check in `McpSpecificationGenerator.methodCall()`:

1. **At tool registration time**: Inspects the `@RequestMapping` annotation on the generated API interface method to determine if the operation is a write (POST/PUT/PATCH/DELETE) vs read (GET/OPTIONS/HEAD)
2. **At tool call time**: For write operations, extracts `clusterName` from MCP arguments and checks `ClustersStorage.getClusterByName(clusterName).isReadOnly()`. If the cluster is read-only, returns a `ReadOnlyModeException` error result instead of invoking the controller method.
3. **Safe operations exempted**: `analyzeTopic`, `cancelTopicAnalysis`, and `registerFilter` are exempted from the readOnly check, consistent with `ReadOnlyModeFilter.SAFE_ENDPOINTS`.

### Changes

- `McpSpecificationGenerator.java`: Inject `ClustersStorage`, add `isWriteOperation()` detection via `@RequestMapping`, add readOnly guard in `methodCall()`
- `McpSpecificationGeneratorTest.java`: Add tests verifying write operations are blocked on readOnly clusters and allowed on non-readOnly clusters

## Test plan

- [x] Existing `testConvertController` test passes (tool generation unchanged)
- [x] New `writeOperationBlockedOnReadOnlyCluster` test: verifies `createTopic` (POST) returns "read-only" error on readOnly cluster
- [x] New `writeOperationNotBlockedOnNonReadOnlyCluster` test: verifies `createTopic` proceeds past readOnly check on normal cluster

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Write operations are now detected and blocked on read-only clusters, returning a clear read-only error; invoking a write without specifying the target cluster now returns a validation error.

* **Tests**
  * Added tests covering read-only enforcement, safe/read operations, behavior for unknown clusters, and validation for missing clusterName.

### DIFF
--- a/api/src/main/java/io/kafbat/ui/service/mcp/McpSpecificationGenerator.java
+++ b/api/src/main/java/io/kafbat/ui/service/mcp/McpSpecificationGenerator.java
@@ -95,7 +95,9 @@ public class McpSpecificationGenerator {
             if (readOnly) {
               return Mono.just(toErrorResult(new ReadOnlyModeException()));
             }
-          } else if (clusterName != null) {
+          } else if (clusterName == null) {
+            return Mono.just(toErrorResult("clusterName is required for write operations"));
+          } else {
             log.warn("Write operation called with non-String clusterName: {}", clusterName.getClass());
           }
         }
--- a/api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
+++ b/api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
@@ -241,6 +241,26 @@ class McpSpecificationGeneratorTest {
         .verifyComplete();
   }
 
+  @Test
+  void writeOperationWithoutClusterNameReturnsError() {
+    ClustersStorage storage = mock(ClustersStorage.class);
+    McpSpecificationGenerator generator = generatorWith(storage);
+    List<AsyncToolSpecification> specs = generator.convertTool(topicsController());
+
+    AsyncToolSpecification createTopic = findTool(specs, "createTopic");
+
+    // Invoke without clusterName in args
+    Mono<CallToolResult> result = invokeTool(createTopic, Map.of());
+
+    StepVerifier.create(result)
+        .assertNext(callResult -> {
+          assertThat(callResult.isError()).isTrue();
+          String text = ((McpSchema.TextContent) callResult.content().get(0)).text();
+          assertThat(text).contains("clusterName is required");
+        })
+        .verifyComplete();
+  }
+
   // --- helpers ---
 
   private static ClustersStorage readOnlyClusterStorage() {


paikend:
Closing in favor of #1766. Good luck @MukulGhare! 🙌

PR #1770, opened 2026-03-30
Infra: Add frontend dependency CVE scanning to CI workflows

Haarolean:

Added frontend dependency vulnerability scanning in two places:
- `pnpm audit` step in the frontend tests workflow for fast, lightweight checks on every PR
- Trivy filesystem scan on `frontend/` in the CVE checks workflow, alongside the existing Docker image scan

- [ ] No need to
- [ ] Unit checks
- [ ] Integration checks

## Summary by CodeRabbit

* **Chores**
  * Added frontend dependency vulnerability scanning in CI to detect CRITICAL and HIGH issues.
  * Added a production-only dependency audit step that fails on high-severity findings.
  * Ensured container-image and filesystem dependency scans run in sequence during CI.
  * Extended CI change-detection to trigger builds when workflow-related frontend files are modified.

### DIFF
--- a/.github/workflows/frontend_pr.yml
+++ b/.github/workflows/frontend_pr.yml
@@ -25,6 +25,7 @@ jobs:
             frontend:
               - "contract/**"
               - "frontend/**"
+              - ".github/workflows/frontend*"
 
   build-and-test:
     needs: changes


.github/workflows/frontend_tests.yml
```diff
@@ -44,6 +44,11 @@ jobs:
           cd frontend/
           pnpm lint:CI
 
+      - name: Audit dependencies
+        run: |
+          cd frontend/
+          pnpm audit --prod --audit-level=high
```
yeikel, line 50: This is not running in the PR due to the `dorny/paths-filter` filters but it seems correct

As a follow up, I think that we should consider updating the filters to also include the workflow folder

Haarolean: >As a follow up, I think that we should consider updating the filters to also include the workflow folder

why? what part of workflow change can affect the dependency list in frontend?

yeikel: Not specifically because it can affect the dependency list, but if a workflow change affects the build (API or frontend or whatever), we should validate it at the time of the change rather than waiting until it reaches main 😄

Haarolean: A workflow change can't affect the dependency list or audit results as those are determined by the lockfile. But it can affect the build environment like node/pnpm version bump, so I added frontend workflows to the path filter.

Issue #1027, opened 2025-04-15
Capture request headers in the audit topic


yeikel:
### Issue submitter TODO list

- [x] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [x] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md) and the feature is not present there

### Is your proposal related to a problem?

I expose the UI behind a proxy that receives and forwards the request.This proxy includes additional and meaningful headers in the request that are currently not captured. 

That causes the following data to be persisted in the audit logs that is not enough to : 

```json

{
	"timestamp": "2025-04-15T20:05:09.872497460Z",
	"username": "Unknown",
	"clusterName": "kafka-cluster",
	"resources": [],
	"operation": "getTopics",
	"result": {
		"success": true
	}
}

```

### Describe the feature you're interested in

Add a new entry in the audit log `requestHeaders` to store a key-value pair with all the headers received

```json

{
	"timestamp": "2025-04-15T20:05:09.872497460Z",
	"username": "Unknown",
	"clusterName": "kafka-cluster",
	"resources": [],
        "requestHeaders": {
         "myHeader" : "myValue"
}, 
	"operation": "getTopics",
	"result": {
		"success": true
	}
}

```

### Describe alternatives you've considered

I considered to this audit at my proxy but it will split the audit to two different places 

### Version you're running

b71a75371c3a75a4ed4df6772744f53e0f0f86af

yeikel:
@Haarolean What are your thoughts about this one? I'd be interested to implement it if accepted

germanosin:
@yeikel Thanks again for your contribution! Appreciate the feature description. Just one small addition from my side — let's include these headers only if they’re explicitly enabled in the config and specified as a list.

yeikel:
> [@yeikel](https://github.com/yeikel) Thanks again for your contribution! Appreciate the feature description. Just one small addition from my side — let's include these headers only if they’re explicitly enabled in the config and specified as a list.

Thank you for the feedback @germanosin 

I updated the feature to reflect this changes. Could you please re-review and share any other feedback? 

Also, please confirm if I can pick this up to send a PR :)

Haarolean:
> [@Haarolean](https://github.com/Haarolean) What are your thoughts about this one? I'd be interested to implement it if accepted

sure thing!

bhasinsaurav:
[...]

PR #1813, opened 2026-04-23
feat: add per-cluster enableMessageViewing config flag

lrudolph333:
Hello, human typing here. Long story short, we are interested in blocking message viewing abilities for all users of our kafbat deployment without any auth requirement. Currently message viewing is only blockable if some sort of RBAC auth is configured, which would introduce unneeded complexity for our use case of having a read only UI for kafka health and metadata. Other users seem to have this use case as well, referenced in the motivation section. 

Claude and I introduce a per cluster environment variable to altogether block message viewing via the UI and any endpoints. Below is an in-depth description, and I've also asked Claude to create an adversarial review of this PR to address potential weak points. Open to discussion on those or other issues, but didn't want to expand the scope of this change too much in one PR. 

Fixes issue #1815 

## Summary

Adds a per-cluster configuration property `enableMessageViewing` (default `true`) that blocks all message operations (read, produce, delete) and KSQL execution without requiring OAuth or LDAP authentication.

### Backend
- `enableMessageViewing` config flag on `ClustersProperties.Cluster` (default `true`)
- `validateMessageViewingEnabled` in `AbstractController` — shared by `MessagesController` and `KsqlController`
- Guards on 7 endpoints: `getTopicMessagesV2`, `sendTopicMessages`, `deleteTopicMessages`, `getSerdes`, `registerFilter` (Messages) + `executeKsql`, `openKsqlResponsePipe` (KSQL)
- Custom `MessageViewingDisabledException` (extends `CustomBaseException`) returns 403 JSON without triggering Spring Security's basic auth popup
- KSQL metadata endpoints (`listStreams`, `listTables`) remain accessible

### Frontend
- `enableMessageViewing` exposed on `ClusterContext` (default `true`)
- When `false`: Messages tab hidden, Messages route removed, Produce Message button hidden, SendMessage sidebar hidden
- Clear Messages buttons hidden in 4 locations: Topic detail dropdown, Topics list ActionsCell, Overview ActionsCell, BatchActionsBar purge button

### API / Config
- OpenAPI spec and TypeSpec contract updated with `enableMessageViewing` field on Cluster model
- Environment variable: `KAFKA_CLUSTERS_0_ENABLEMESSAGEVIEWING=false`

## Motivation

Many kafbat deployments run with `auth.type: DISABLED` and have no OAuth/LDAP provider configured. These deployments currently have **no way to restrict message viewing** — the existing RBAC system (`MESSAGES_READ` permission) only works when authentication is enabled, because it needs a user identity to match against role subjects.

This was originally requested in [kafbat/kafka-ui#1027](https://github.com/provectus/kafka-ui/issues/1027), where users asked for a simple startup flag to disable message viewing. The issue was closed when RBAC was implemented via [kafbat/kafka-ui#753](https://github.com/provectus/kafka-ui/issues/753), but RBAC requires authentication — which doesn't help deployments that can't or don't want to set up an identity provider.

Key quotes from the original discussion:
- *"Some of our data has restricted access. Currently this tool will allow anyone to browse this data."* — @akamensky
- *"The fact that some of data should not be available to user to read at all (security concern)... just an option to completely disable reading messages via UI would be awesome."* — @akamensky
- *"Users are trying to live-tail topics with ~2mm msgs/s which causes all sorts of issues"* — @akamensky

## Configuration

```yaml
kafka:
  clusters:
    - name: my-cluster
      bootstrapServers: localhost:9092
      enableMessageViewing: false   # blocks all message operations for this cluster
```

Or via environment variable:
```
KAFKA_CLUSTERS_0_ENABLEMESSAGEVIEWING=false
```

## Bugs fixed along the way

1. **Basic auth popup on 403** — The original implementation threw `AccessDeniedException`, which Spring Security intercepted to produce a 401 + `WWW-Authenticate: Basic` header instead of a clean 403 JSON response. Fixed by introducing `MessageViewingDisabledException extends CustomBaseException`, which routes through `GlobalErrorWebExceptionHandler`.

2. **Access Denied page blocking all topic operations** — The `SendMessage` component mounted unconditionally inside `SlidingSidebar`, calling `useSerdes()` via `useSuspenseQuery`. When message viewing was disabled, the 403 from the serdes endpoint threw through the error boundary and replaced the entire topic page with an Access Denied page. Fixed by wrapping the sidebar in an `enableMessageViewing` guard.

3. **KSQL bypass** — `KsqlController.executeKsql()` and `openKsqlResponsePipe()` allowed reading messages via `PRINT <topic>` or `SELECT * FROM stream` without checking the flag. Fixed by adding `validateMessageViewingEnabled` calls to both endpoints.

### Backend integration tests (`MessagesControllerDisableViewingTest`)
- `getTopicMessagesV2` → 403 without `WWW-Authenticate` header
- `getSerdes` → 403
- `sendTopicMessages` → 403
- `deleteTopicMessages` → 403
- `registerFilter` → 403
- Serdes on non-disabled cluster → 200 (positive control)

### Backend integration tests (`KsqlControllerDisableViewingTest`)
- `executeKsql` → 403
- `openKsqlResponsePipe` → 403

### Frontend tests (`Topic.spec.tsx`)
- Messages tab not rendered when `enableMessageViewing=false`
- Messages route not rendered on direct navigation
- Other tabs (Overview, Consumers, Settings) still render
- Produce Message button not rendered
- SendMessage sidebar not rendered
- Clear messages dropdown item not rendered

All tests follow the existing pattern: a dedicated `messageViewingDisabledLocal` test cluster in `AbstractIntegrationTest` with `enableMessageViewing=false`.

## Test plan
- [x] Configure cluster with `ENABLEMESSAGEVIEWING=false` and verify Messages tab is hidden (see screenshots below)
- [x] Verify Produce Message button and SendMessage sidebar are hidden
- [x] Verify Clear Messages buttons are hidden in all 4 locations
- [x] Verify all message-related API endpoints return 403 JSON (no basic auth popup)
- [x] Verify KSQL execute/response endpoints return 403
- [x] Verify KSQL metadata endpoints (listStreams, listTables) still work
- [x] Verify clusters without the flag continue to work normally
- [x] Verify no Access Denied page when navigating topic pages
- [x] Backend integration tests pass
- [x] Frontend unit tests pass

After Screenshots
Screenshot 2026-04-22 at 11 27 58 AM
Screenshot 2026-04-22 at 11 27 29 AM

Before screenshots:
Screenshot 2026-04-22 at 11 28 10 AM
Screenshot 2026-04-22 at 11 27 43 AM

Screenshots of message disablement using RBAC approach:
Screenshot 2026-04-23 at 5 18 47 PM
Screenshot 2026-04-23 at 5 18 39 PM

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added cluster-level configuration property to disable message viewing functionality
  * When disabled, the Messages tab, message production, and clear-message actions are hidden from the UI
  * Message-related API calls return a forbidden error when message viewing is disabled for a cluster
  * Feature is enabled by default for backward compatibility


lrudolph333:
@Haarolean @germanosin I see you've both been pretty active with the files I'm touching here, I would be curious to hear your thoughts on this issue/approach.

germanosin:
> @Haarolean @germanosin I see you've both been pretty active with the files I'm touching here, I would be curious to hear your thoughts on this issue/approach.

Hi, to be honest this feature doesn’t feel like the right approach from my perspective. Setting up RBAC isn’t as complex as it might seem, so I’d suggest giving it a try first.

Haarolean:
Agreed with German, today it's "no messages w/o RBAC", tomorrow "no X without RBAC" and the next thing we know, we have to support yet another RBAC built upon no RBAC/auth.

lrudolph333:
That makes sense, I'll revisit how we can use RBAC auth with our setup.

Issue #1630, opened 2026-01-09
Metrics tab of MSK broker results in a 404 error instead of metrics gathered from open monitoring endpoint


hogarth-sv:
### Issue submitter TODO list

- [x] I've looked up my issue in [FAQ](https://ui.docs.kafbat.io/faq/common-problems)
- [x] I've searched for an already existing issues [here](https://github.com/kafbat/kafka-ui/issues)
- [x] I've tried running `main`-labeled docker image and the issue still persists there
- [x] I'm running a supported version of the application which is listed [here](https://github.com/kafbat/kafka-ui/blob/main/.github/SECURITY.md)

### Describe the bug (actual behavior)

MSK cluster configured with Open Monitoring enabled.

Kafbat deployed in an EKS cluster with a config that works to see brokers, topics, consumers, etc etc 

From the pod I can use wget and see the prometheus data from http://brokernode:11001/metrics

The logs from kafbat do not show any errors getting metrics:
```
2026-01-09T19:08:51.801Z DEBUG 1 --- [or-http-epoll-6] i.k.u.s.ClustersStatisticsScheduler      : Metrics updated for cluster: cloud
2026-01-09T19:09:07.124Z DEBUG 1 --- [     parallel-8] i.k.u.s.m.s.i.InferredMetricsScraper     : 16 metric families inferred from cluster state
2026-01-09T19:09:07.124Z DEBUG 1 --- [     parallel-8] i.k.u.s.ClustersStatisticsScheduler      : Metrics updated for cluster: onprem
2026-01-09T19:09:18.878Z DEBUG 1 --- [    parallel-13] i.k.u.s.ClustersStatisticsScheduler      : Start getting metrics for kafkaCluster: cloud
2026-01-09T19:09:18.878Z DEBUG 1 --- [    parallel-14] i.k.u.s.ClustersStatisticsScheduler      : Start getting metrics for kafkaCluster: onprem
2026-01-09T19:09:19.159Z DEBUG 1 --- [    parallel-16] i.k.u.s.m.s.i.InferredMetricsScraper     : 18 metric families inferred from cluster state
2026-01-09T19:09:19.160Z DEBUG 1 --- [    parallel-16] i.k.u.s.m.s.p.PrometheusMetricsRetriever : Retrieving metrics from prometheus endpoint: broker-1:11001
[...]

PR #1821, opened 2026-04-28
BE: Drop non-finite metric data points instead of failing the broker metrics response

### LINKED ISSUE #1820: Broker metrics endpoint returns 404 when any data point is NaN/Infinity
### Describe the bug

`GET /api/clusters/{clusterName}/brokers/{id}/metrics` returns **HTTP 404** whenever the broker's Prometheus exposition contains even a single `NaN` (or `Infinity`) value, instead of returning the metrics that *are* available.

The frontend then shows the broker's *Metrics* tab as empty / errored.

This is the same outward symptom as #1630 — but with a deterministic, easily reproducible root cause that #1630 did not pin down. (#1630 also reports the silent variant where the response body is empty with HTTP 200; the underlying mapping path is the same.)

### Root cause

`io.kafbat.ui.mapper.ClusterMapper#convert(Stream<MetricSnapshot>)` does:

```java
.value(BigDecimal.valueOf(readPointValue(p)))
```

`BigDecimal.valueOf(double)` throws `NumberFormatException("Infinite or NaN")` on `Double.NaN` / `Double.POSITIVE_INFINITY` / `Double.NEGATIVE_INFINITY`.

`BrokersController#getBrokersMetrics` catches **all** errors with:

```java
.onErrorReturn(ResponseEntity.notFound().build())
```

So a single NaN data point anywhere in the broker's metric stream collapses the whole response to 404. Nothing is logged.

### How NaN ends up in the exposition

JMX-Prometheus exporter (Strimzi's `metricsConfig.type: jmxPrometheusExporter`, MSK Open Monitoring, anything using `io.prometheus.jmx`) legitimately emits `NaN` for `*_avg` / `*_max` Kafka sensors when the underlying meter has never been hit. Concretely, on a fresh Strimzi broker we see ~24 NaN data points like:

```
kafka_server_socket_server_metrics_reauthentication_latency_avg{listener="PLAIN-9092",networkProcessor="3"} NaN
kafka_server_socket_server_metrics_reauthentication_latency_max{listener="REPLICATION-9091",networkProcessor="0"} NaN
kafka_server_socket_server_metrics_request_size_avg{listener="TLS-9093",networkProcessor="6"} NaN
...
```

Operators have no realistic way to make these non-NaN — the broker has simply never observed a reauthentication or a request on that listener. Today, that means the broker's Metrics tab never works.

### Steps to reproduce

1. Run a Kafka cluster whose Prometheus endpoint produces at least one `NaN` data point. The simplest reproduction is Strimzi with `jmxPrometheusExporter` and the default JMX ruleset, but any cluster with unused listeners/sensors works.
2. Configure kafbat-ui:
   ```yaml
   KAFKA_CLUSTERS_0_METRICS_TYPE: PROMETHEUS
   KAFKA_CLUSTERS_0_METRICS_PORT: 9404   # 11001 for MSK Open Monitoring
   ```
3. Verify the endpoint serves data:
   ```
   wget -qO- http://broker-0:9404/metrics | grep -c ' NaN$'
   # > 0
   ```
4. Open the broker detail → Metrics tab, or curl `/api/clusters/<name>/brokers/0/metrics`.

### Expected behavior

Finite data points are returned; NaN/Infinity points are dropped (or otherwise represented in a JSON-safe way). A single bad point should not nuke the whole response.

### Actual behavior

`HTTP 404 Not Found` (or empty body in some Spring/Reactor configurations — see #1630). No log line.

### Environment

- kafbat-ui: `ghcr.io/kafbat/kafka-ui:v1.5.0` (also reproducible against `main` — same `ClusterMapper` code path)
- Kafka: Strimzi 0.x, `jmxPrometheusExporter`, KRaft
- The same code path is hit by MSK Open Monitoring (port 11001) — see #1630.

### Fix

PR will follow this issue. The minimal fix is to filter non-finite data points in `ClusterMapper#convert` before they reach `BigDecimal.valueOf`. A separate concern — the silent `onErrorReturn(notFound())` in `BrokersController` swallowing all errors — is intentionally left out of scope for this issue/PR.

titaniper:

Fixes #1820 (related: #1630).

`ClusterMapper#convert` builds `MetricDTO`s by calling `BigDecimal.valueOf(readPointValue(p))`. `BigDecimal.valueOf(double)` throws `NumberFormatException("Infinite or NaN")` on any non-finite input. JMX-Prometheus exporters (Strimzi `jmxPrometheusExporter`, MSK Open Monitoring, plain `io.prometheus.jmx`) legitimately emit `NaN` for `*_avg` / `*_max` Kafka sensors that have never been observed (e.g. reauthentication latency on a never-used listener, request size on an idle replication listener).

The exception bubbles up to `BrokersController#getBrokersMetrics`, which silently swallows it via `.onErrorReturn(ResponseEntity.notFound().build())`. Net effect: a single NaN data point anywhere in the broker's metric stream collapses the entire `/api/clusters/{name}/brokers/{id}/metrics` response to **404**, with no log line. The broker's Metrics tab never works.

This change makes `convert()` filter non-finite data points before they reach `BigDecimal.valueOf`. Finite points still flow through; the broken point is dropped rather than poisoning the whole response.

The silent `.onErrorReturn(notFound())` in `BrokersController` is a separate issue (it hides every other failure mode in this path too) and is intentionally left out of scope here.

- Is dropping the offending point the right semantic? Alternatives considered:
  - Substitute `0` (changes semantics — `0` is a valid value).
  - Substitute `null` in `MetricDTO.value` (the OpenAPI contract has it as a non-nullable BigDecimal-shaped string in JSON; changing that ripples to clients).
  - Sanitize at the scrape layer in `PrometheusTextFormatParser` (broader blast radius — affects the global `/metrics` re-export too, where Prometheus itself accepts `NaN` and would arguably want it preserved).
  Filtering at the DTO conversion layer keeps the fix localized to the API surface that breaks today.
- The same `convert()` is also called from `toClusterMetrics` (cluster summary). Today that path tends not to hit NaN because `SummarizedMetrics` aggregates, but it's still worth confirming the filter doesn't drop anything operators rely on. I don't think it does — a `NaN`-aggregated bucket would be unusable anyway.

- [ ] No need to
- [x] Unit checks
- [ ] Integration checks

- New unit test `ClusterMapperTest#toBrokerMetricsDropsNonFiniteDataPoints` builds a `BrokerMetricsDTO` from four `GaugeSnapshot`s (NaN, +Inf, -Inf, finite) and asserts only the finite point survives.
- Reproduced manually against a Strimzi 0.x broker on stage: with stock kafbat-ui v1.5.0 the endpoint returned 404; with this patch it returns 200 with the broker's metrics minus the ~24 NaN points.

- [ ] I have made corresponding changes to the documentation (e.g. **ENVIRONMENT VARIABLES**) — n/a, no user-facing config change

🐢 (a tortoise — slow, finite, never NaN)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced broker metrics processing to handle invalid numerical values more robustly. Metric values that are non-finite (such as NaN or Infinity) from Prometheus data sources are now properly filtered out during conversion, preventing collection errors and ensuring reliable, continuous metrics monitoring and reporting throughout the system.

### DIFF
--- a/api/src/main/java/io/kafbat/ui/mapper/ClusterMapper.java
+++ b/api/src/main/java/io/kafbat/ui/mapper/ClusterMapper.java
@@ -71,12 +71,22 @@ public interface ClusterMapper {
     return metrics
         .flatMap(m ->
             m.getDataPoints().stream()
-                .map(p ->
-                        new MetricDTO()
-                            .name(m.getMetadata().getName())
-                            .labels(p.getLabels().stream().collect(toMap(Label::getName, Label::getValue)))
-                            .value(BigDecimal.valueOf(readPointValue(p)))
-                )
+                .flatMap(p -> {
+                  double value = readPointValue(p);
+                  // BigDecimal.valueOf(double) throws NumberFormatException on NaN/Infinity.
+                  // JMX/Prometheus exporters legitimately emit NaN for unused metrics
+                  // (e.g. average latency on a never-used listener). Drop those points
+                  // instead of failing the whole broker metrics response.
+                  if (!Double.isFinite(value)) {
+                    return Stream.empty();
+                  }
+                  return Stream.of(
+                      new MetricDTO()
+                          .name(m.getMetadata().getName())
+                          .labels(p.getLabels().stream().collect(toMap(Label::getName, Label::getValue)))
+                          .value(BigDecimal.valueOf(value))
+                  );
+                })
         );
   }
 
--- a/api/src/test/java/io/kafbat/ui/mapper/ClusterMapperTest.java
+++ b/api/src/test/java/io/kafbat/ui/mapper/ClusterMapperTest.java
@@ -1,0 +1,39 @@
+package io.kafbat.ui.mapper;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.kafbat.ui.model.BrokerMetricsDTO;
+import io.prometheus.metrics.model.snapshots.GaugeSnapshot;
+import io.prometheus.metrics.model.snapshots.Labels;
+import io.prometheus.metrics.model.snapshots.MetricMetadata;
+import io.prometheus.metrics.model.snapshots.MetricSnapshot;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class ClusterMapperTest {
+
+  // Reproduces https://github.com/kafbat/kafka-ui/issues/1630 — JMX/Prometheus exporters
+  // emit NaN for unused gauges (e.g. average latency on a never-used listener). Without
+  // a guard, BigDecimal.valueOf(NaN) throws NumberFormatException and the whole broker
+  // metrics response is silently turned into 404 by BrokersController.
+  @Test
+  void toBrokerMetricsDropsNonFiniteDataPoints() {
+    ClusterMapper mapper = new ClusterMapperImpl();
+
+    var withNan = new GaugeSnapshot(new MetricMetadata("nan_metric", "help"), List.of(
+        new GaugeSnapshot.GaugeDataPointSnapshot(Double.NaN, Labels.of("l", "v"), null)));
+    var withPosInf = new GaugeSnapshot(new MetricMetadata("posinf_metric", "help"), List.of(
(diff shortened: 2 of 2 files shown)


Haarolean (approved):

Haarolean:
@titaniper thanks for your first contribution to kafbat ui. This will be merged soon

PR #1842, opened 2026-05-08
BE: Restore SchemaRegistry auto-selection for Avro topics (#1833)

### LINKED ISSUE #1833: Regression: version 1.5.0 does not auto‑select SchemaRegistry value serde for Avro topics
### Issue submitter TODO list

In version 1.5.0, Kafka topics containing Avro‑serialized messages are not automatically deserialized in the UI. The `String` value serde is selected by default, resulting in raw Avro serialized messages being displayed. This is a regression from previous versions, where the `SchemaRegistry` value serde was automatically selected. In version 1.5.0, users must now manually select `SchemaRegistry` to view the content of Avro serialized messages.

### Expected behavior

I expect the `SchemaRegistry` value serde to be automatically selected for topics containing Avro serialized messages.

### Your installation details

- _App version:_ afc9c91 (v1.5.0)
- _Helm chart version_: 1.6.3

### Steps to reproduce

- Deploy Kafka cluster and Kafka schema registry, and create a topic with Avro serialized messages.
- Deploy Kafka UI using Helm chart version 1.6.3.
- Observe that Kafka UI automatically selects the `String` value serde for the Kafka topic.

### Screenshots

Example of Kafka topic with Avro serialized messages in Kafka UI version 1.5.0.
The `String` value serde is selected by default, and the raw Avro message is displayed to the user:

Image

Example of Kafka topic with Avro serialized messages in Kafka UI version 1.4.2.
The `SchemaRegistry` value serde is selected by default, and the deserialized Avro message is displayed to the user:

Image

### Logs

_No response_

### Additional context

_No response_

joshuaNathaniel:
Fixes #1833

In 1.5.0 the value serde stopped auto-selecting SchemaRegistry for Avro topics and defaulted to String. This brings back the 1.4.2 behavior.

The fix is in three places:

1. `ClusterSerdes.findSerdeByPatternsOrDefault` now runs three passes: explicit `topicKeysPattern`/`topicValuesPattern` first, then `defaultKeySerde`/`defaultValueSerde`, then any serde that says it should be preferred via `couldBePreferable` and passes `canSerialize`/`canDeserialize`. Previously `couldBePreferable` was wired as a `continue` inside the pattern loop, so it never actually selected anything (auto-configured SR has a null pattern), and as a side effect it could silently skip explicit user patterns too.

2. `SchemaRegistrySerde.couldBePreferable` now matches both TopicNameStrategy and TopicRecordNameStrategy. RecordNameStrategy is left out because subjects there are bare FQNs with no relationship to the topic, so we can't safely guess which topics use them. SR still shows up in the dropdown for those topics (via `canSerialize`), it's just not the default.

3. Default for `Serde#couldBePreferable` flipped from `true` to `false`. With pass 3 in place, leaving it at `true` would have made `StringSerde` win for everything (it's registered first and `canDeserialize` is unconditionally true). The opt-in default also matches what `SchemaRegistrySerde` was already doing as the only override.

Tests:

- `ClusterSerdesTest`: there wasn't any coverage on the selection logic, which is why this regression slipped through. Pins the three-pass behavior and the original bug.
- `SchemaRegistrySerdeTest.CouldBePreferableTests`: covers each naming strategy plus a few edge cases (opposite target, mixed registry state, subjects belonging to other topics).
- `SerdesInitializerTest`: assertion that `defaultValueSerde` stays null when not configured, so nobody tries to "fix" things by re-adding the old implicit fallback.

- The pass ordering in `findSerdeByPatternsOrDefault`. Want to make sure user config still wins over implicit detection.
- The choice to keep `couldBePreferable` narrower than `canSerialize`. Comment explains it but worth a second pair of eyes.
- The `couldBePreferable` interface default flip. It's in `serde-api` so technically a contract change, but in 1.5.0 the default was effectively unused (the only consumer was the dead `continue` inside pass 1), so anyone upgrading without overriding the method should see no behavioral difference.

**How Has This Been Tested?**

  - Walked through the subject-strategy-demo for all three strategies across Avro/Protobuf/JSON. Topic-affiliated strategies default to `SchemaRegistry`; RecordNameStrategy defaults to `String`.
- [x] Unit checks
- [x] Integration checks

**Checklist**

fennec-fox

## Summary by CodeRabbit

* **New Features**
  * Improved serde selection so the app now more reliably picks the most relevant serializer/deserializer for a topic, with clearer fallback behavior.
  * Added support for additional OAuth-related configuration options, including client authentication settings and schema registry OAuth configuration.

* **Bug Fixes**
  * Fixed cases where the wrong serde could be chosen when multiple candidates matched.
  * Improved schema registry detection so topic-specific and record-name subjects are recognized more accurately.

* **Tests**
  * Added coverage for serde selection precedence and schema registry matching scenarios.


api/src/main/resources/static/openapi/kafbat-ui-api.yaml
joshuaNathaniel, line 1: @Haarolean This is a generated file. Do you even want me committing this? I was debating whether I should include it in the commit. 😬

joshuaNathaniel: Actually, I'm going to remove for now.

Haarolean: no, why? we have legacy openapi file that shouldn't be updated and new typespec ones that deterministically generate a new openapi one, there's no point in committing one.

IversenMichael:
@joshuaNathaniel Thank you for working on the issue! In version 1.4.2, the SchemaRegistry value serde was automatically selected even for topics using the RecordNameStrategy where the schema must be inferred from the messages themselves and not only from the topic name. Will this also be the case in version 1.5.0 with this fix?

joshuaNathaniel:
> @joshuaNathaniel Thank you for working on the issue! In version 1.4.2, the SchemaRegistry value serde was automatically selected even for topics using the RecordNameStrategy where the schema must be inferred from the messages themselves and not only from the topic name. Will this also be the case in version 1.5.0 with this fix?

This is a good callout @IversenMichael. Unfortunately, no, it does not. This is why I am looking for feedback on it. I believe the prior behavior was to always default topics using key and value serde to the schema registry for every topic.

I think the intention was to fix the lying dropdown before and we used couldBePreferable to only nominate SR when there were actually linked subjects. But I think there is a bit of a performance concern mixed in here for RecordNameStrategy. 

Yeah, I made a comment in that issue and I deleted it. I'm starting to regret that. I think @germanosin has to weigh in on this. We can certainly restore the behavior but how we do it has trade offs.

joshuaNathaniel:
Popping back in here. It's been a few days. I just need a wee bit of direction on where we want to go in regards to my last comment. I'm more than happy to do the work and make everyone happy and whole again. :D

joshuaNathaniel:
Reworked serde auto-selection so SchemaRegistry is picked as the default again when a topic has schemas, fixing the regression where Avro topics fell back to String. ClusterSerdes now resolves in three ordered passes: explicit topic patterns, configured cluster defaults, then auto-detection for serdes without a pattern. couldBePreferable on the Serde interface now defaults to false so serdes opt in, and SchemaRegistrySerde opts in when the registry holds a subject the topic could be using, covering all three naming strategies: TopicNameStrategy and TopicRecordNameStrategy are scoped to the specific topic, while bare RecordNameStrategy subjects apply to any topic since they can't be tied to one. This keeps @Haarolean's fix for explicitly configured serdes without patterns intact. Added tests across ClusterSerdesTest, SchemaRegistrySerdeTest, and SerdesInitializerTest covering each strategy, the three-pass precedence, and sibling-topic prefix cases.

api/src/main/java/io/kafbat/ui/serdes/builtin/sr/SchemaRegistrySerde.java
```diff
@@ -398,7 +403,36 @@ public List<SerdeParameter> getParameters(String topic, Target type) {
...
+    // magic-byte check decodes real SR messages and the fallback serde handles the rest.
+    String defaultSubject = schemaSubject(topic, type);
+    String topicPrefix = topic + "-";
+    return getAllSubjects().stream().anyMatch(subject -> {
+      if (subject.equals(defaultSubject)) {
+        return true;
+      }
+      if (subject.indexOf('-') < 0) {
+        // RecordNameStrategy: bare record name, not tied to any topic.
+        return true;
```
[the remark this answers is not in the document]
joshuaNathaniel: This accepts that if record name strategy subject is used within the schema registry, then all topics would need to prefer SchemaRegistry. There is no way around this.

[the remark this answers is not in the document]
joshuaNathaniel: The preference is to do this correctly. The intended behavior is confirmed.

api/src/main/resources/static/openapi/kafbat-ui-api.yaml
```diff
@@ -3306,6 +3308,29 @@ components:
                             type: string
                           password:
                             type: string
+                          oauth:
```
Haarolean, line 3311: unrelated

Haarolean (approved):

api/src/main/resources/static/openapi/kafbat-ui-api.yaml
Haarolean: @joshuaNathaniel can you revert please?

joshuaNathaniel:
@Haarolean What do you need to me to do here. Just get this back to good standing? I see there was the kafbat-ui-api.yaml change.

PR #1928, opened 2026-08-06
Prevent ClustersStatisticsScheduler from failing all clusters on single error

### LINKED ISSUE #1638: Multiple clusters failing the UI if one has a bad config
### Issue submitter TODO list

Hi I have installed kafka-ui v1.4.1 and I am trying to create multiple clusters.
If one of my cluster has a bad config, for example a bad password both clusters fails and the one that is correctly set up never becomes green.
Is this a known issue ? how can I fix this ? 
And yes I can make sure that a password is never wrong and what not ... but passwords can expire i'd like to avoid it crashing because of one bad cluster in my production. 

### Expected behavior

Kakfa-ui runs and only the cluster with the bad config is not ready

### Your installation details

1. v1.4.1
2. helm latest one
3. two cluster configured via configmap
4. Running on eks 1.31

### Steps to reproduce

just configure two cluster, one of which has some bad config, bad  bootstrap url or password

### Screenshots

_No response_

### Logs

_No response_

### Additional context

_No response_

jakub-blazej-telekom:

Fixes #1638.

`ClustersStatisticsScheduler.updateStatistics` collapses all configured clusters
into a single `Flux.parallel().flatMap(updateCache).then().block()` chain. When
any single cluster's `updateCache` signals `onError`, the parallel `Flux`
terminates and cancels the sibling inner `Mono`s, so no cluster in that tick —
healthy or not — gets its cache updated. Every subsequent tick hits the same
error path and every cluster is left stuck at `INITIALIZING`.

Reproduced against a real deployment with 4 healthy SASL/SCRAM clusters + 1
with bad credentials: all 5 remain `INITIALIZING` and the UI is unusable.
Removing only the broken cluster restores the other 4.

Wrap each per-cluster update in `.onErrorResume(...)` so a failing cluster only
logs and drops out of that tick, leaving the rest to complete. After the fix
the 4 healthy clusters reach `ONLINE` and the broken one reports `OFFLINE`.

Whether the log-and-swallow policy at the scheduler layer is the right level —
`StatisticsService.getStatistics` already has an `.onErrorResume(t -> Mono.just(Statistics.statsUpdateError(t)))`
that converts most failures to an `OFFLINE` cache entry, but some paths still
propagate through `.block()` and abort the parallel `Flux`. The scheduler-level
guard makes the scheduler itself resilient to any such leak.

**How Has This Been Tested?**

- [ ] No need to
- [x] Unit checks
- [ ] Integration checks

Manual: 4 SASL/SCRAM clusters + 1 broken-cred cluster. Before: all 5
`INITIALIZING`, `/api/clusters/*/brokers` returns 500. After: 4 `ONLINE`,
broken one `OFFLINE`.

Unit: added `ClustersStatisticsSchedulerTest.oneFailingClusterDoesNotBlockOthers`.
Fails on `main` (test throws `IllegalStateException` via `block()`), passes
with this change.

**Checklist**

meerkat

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cluster statistics updates so a failure for one Kafka cluster no longer prevents updates for other clusters.
  * Statistics update errors are handled gracefully, allowing processing to continue.


Haarolean (approved):

Haarolean:
@jakub-blazej-telekom thanks for your contribution!

jakub-blazej-telekom:
I see the 3 of the failing tests are common across other PRs and the different one is 30 seconds of `Connection to node 1 (localhost/127.0.0.1:32777) could not be established. Node may not be available.` and then `InterruptedException`, so I expect it is not caused by the proposed change.

PR #1937, opened 2026-08-14
Allow OAuth2 resource-server auth without requiring a dummy client registration

### LINKED ISSUE #1929: Allow OAUTH2 resource-server auth without requiring dummy client registration

When `auth.type=OAUTH2` is set and the goal is purely bearer/resource-server JWT auth (e.g. exposing `/mcp/sse` to non-interactive service/agent clients, no browser login involved at all), kafka-ui still refuses to start unless at least one `auth.oauth2.client.*` registration is configured too.

`OAuthSecurityConfig.clientRegistrationRepository()`:

```java
if (registrations.isEmpty()) {
  throw new IllegalArgumentException("OAuth2 authentication is enabled but no providers specified.");
}
```

This forces operators who only want machine-to-machine/bearer access to also configure a full, working dummy OAuth2 login client (client-id, secret, authorization/token/userinfo/jwk-set URIs) purely to satisfy this startup check — even though that client is never actually used by anyone.

Allow `auth.type=OAUTH2` to start successfully with only `auth.oauth2.resource-server` configured and zero `auth.oauth2.client.*` entries. The interactive `oauth2Login` part of the filter chain could simply be skipped when there are no client registrations and a resource-server config is present.

Configuring a dummy/unused client registration as a workaround — works, but adds confusing, unused config that looks like a real login provider.

### Version you're running

kafbat/kafka-ui:v1.5.0 and :main

### Additional context

Hit this while setting up bearer-token auth for the MCP server (`/mcp/sse`) for a non-interactive AI-agent use case. Happy to help test a fix.

Andrey834:
### What / why

Currently, when `auth.type=OAUTH2` is set, kafka-ui refuses to start unless at least one `auth.oauth2.client.*` registration is configured, even if the goal is purely bearer/resource-server JWT auth for non-interactive (service/agent) clients. `OAuthSecurityConfig#clientRegistrationRepository()` throws if `registrations.isEmpty()`, regardless of whether a resource-server config is present.

This forces operators who only need machine-to-machine/bearer access to configure a full, unused dummy OAuth2 login client (client-id, secret, authorization/token/userinfo/jwk-set URIs) purely to satisfy this startup check.

### Solution

`clientRegistrationRepository()` now only throws when **both** client registrations and `auth.oauth2.resource-server` are absent (i.e. nothing configured at all). When there are no client registrations but a resource-server *is* configured, the bean is legitimately **absent** (`null`) instead of throwing.

This mirrors a pattern already used elsewhere in the codebase: `AccessControlService` already declares its `ReactiveClientRegistrationRepository` dependency as `@Nullable` and handles a missing repository gracefully. This PR extends that same convention to the two other places in `OAuthSecurityConfig` that assumed a repository always exists:

- `configure(...)`: the interactive `.oauth2Login(...)` / `.logout(...)` filter chain wiring is now skipped entirely when there's no client registration repository, only `.oauth2ResourceServer(...)` is wired for bearer/JWT-only setups.
- `defaultOidcLogoutHandler(...)`: falls back to a no-op `ServerLogoutSuccessHandler` when the repository is absent, instead of failing to construct `OidcClientInitiatedServerLogoutSuccessHandler` (which requires a non-null repository).

No changes were needed in `AccessControlService`, its existing `@Nullable` handling already covers this case correctly.

### Testing

- Added `OAuthResourceServerOnlyTest`: boots `OAuthSecurityConfig` with only `auth.oauth2.resource-server.jwt.jwk-set-uri` set and zero client registrations, asserts the context loads and `ReactiveClientRegistrationRepository` is legitimately absent.
- Ran the full existing `io.kafbat.ui.config.auth.*` test suite (proxy/without-proxy, Azure, Cognito, standard client-registration flow), all passing, no regressions.
- `checkstyleMain` / `checkstyleTest`, clean.

Fixes #1929

## Summary by CodeRabbit

- **New Features**
  - OAuth2 resource-server-only configurations can now start without client registrations.
  - Support includes JWT and opaque-token authentication setups.
  - Interactive OAuth login and OIDC logout are enabled only when OAuth clients are configured.

- **Bug Fixes**
  - Applications using resource-server authentication no longer require unnecessary OAuth client settings.
  - Startup validation remains in place when neither authentication configuration is provided.

### DIFF
--- a/api/src/main/java/io/kafbat/ui/config/auth/OAuthSecurityConfig.java
+++ b/api/src/main/java/io/kafbat/ui/config/auth/OAuthSecurityConfig.java
@@ -114,7 +114,7 @@ public class OAuthSecurityConfig extends AbstractAuthSecurityConfig {
           .logout(spec -> spec.logoutSuccessHandler(logoutHandler));
     }
 
-    if (properties.getResourceServer() != null) {
+    if (isResourceServerUsable(properties.getResourceServer())) {
       OAuth2ResourceServerProperties resourceServer = properties.getResourceServer();
       if (resourceServer.getJwt() != null && resourceServer.getJwt().getJwkSetUri() != null) {
         builder.oauth2ResourceServer(c -> c.jwt(j ->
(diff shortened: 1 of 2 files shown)


Andrey834:
Addressed the merge-risk concern in the latest commit (a310514): `clientRegistrationRepository()` now checks that `resource-server` actually has a usable `jwt.jwk-set-uri` or `opaquetoken.introspection-uri` before treating an empty client list as valid. If `resource-server` is present but effectively empty, it throws a clear `IllegalArgumentException` at startup instead of silently starting with no authentication mechanism wired.

Added unit tests covering: neither client nor resource-server configured, resource-server present but empty, and resource-server with a valid jwt config. Full existing `io.kafbat.ui.config.auth.*` suite and checkstyle still pass.

Andrey834:
Added `returnsNullWhenResourceServerHasOpaqueToken` in the latest commit (75b51f0), mirrors the existing jwt test but for the opaquetoken.introspection-uri branch. Full suite still passing.

PR #1948, opened 2026-08-19
Fix Flux serialization in nested ResponseEntity for MCP tool results

### LINKED ISSUE #1454: MCP server  doesn't properly handle Flux responses.

kanchan78:
## Summary

Fixes #1454.

`McpSpecificationGenerator.toCallResult()` already handles a bare `Flux<?>` correctly by calling `.collectList()` before serializing. But controllers like `ClustersController#getClusters`, `AclsController#listAcls`, `KafkaConnectController#getAllConnectors`, and `MessagesController#getTopicMessagesV2` return `Mono<ResponseEntity<Flux<T>>>` — the `Flux` is nested *inside* the `ResponseEntity` body.

`reponseToCallResult()` unwraps the `ResponseEntity` and passes its body straight to `callToolResult()`, skipping the `Flux`-aware branch in `toCallResult()` entirely. Jackson then introspects the live, unresolved `Flux` object graph and serializes its internal Reactor fields instead of its contents, e.g.:

```json
{"scanAvailable":true,"prefetch":-1}
```

instead of the actual list of clusters/ACLs/connectors/messages. The tool call reports `isError:false`, so an MCP client gets silently wrong data rather than a visible failure.

### Fix

`reponseToCallResult()` now delegates the response body to `toCallResult()` instead of calling `callToolResult()` directly, so the existing `Flux` branch (`flux.collectList()`) applies uniformly whether the `Flux` is top-level or nested inside a `ResponseEntity`.

### Verification

- Reproduced against a live deployment: `getClusters`, `listAcls`, `getAllConnectors`, and `getTopicMessagesV2` all returned the `{"scanAvailable":...}` garbage before the fix; `getTopics`, `getConsumerGroupsPage`, `getSchemas`, and `getLatestSchema` (which return `Page<T>`/plain DTOs, not `Flux`) were unaffected — confirming the bug is specific to the `ResponseEntity<Flux<T>>` shape.
- Added a regression test (`toolBackedByResponseEntityOfFluxReturnsCollectedDataNotReactorInternals`) exercising `ClustersController#getClusters` through the real `McpSpecificationGenerator`. Verified it fails on unpatched code and passes after the fix.
- Built the full application (including frontend) and smoke-tested the resulting image standalone: opened a real MCP SSE session and called `getClusters`, which returned `[]` (correct, no clusters configured) instead of the Reactor-internals payload.

## Test plan

- [x] `./gradlew :api:test --tests "io.kafbat.ui.service.mcp.McpSpecificationGeneratorTest"` — all 8 tests pass (7 pre-existing + 1 new regression test)
- [x] `./gradlew :api:checkstyleMain :api:checkstyleTest` — clean
- [x] Built `api` module + Docker image from source, ran it, confirmed `getClusters` over a live MCP SSE session returns valid JSON instead of Reactor internals

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed MCP tool results backed by reactive responses to return the resolved data correctly.
  * Cluster data from streaming responses is now collected and displayed instead of exposing internal reactive-system details.

### DIFF
--- a/api/src/main/java/io/kafbat/ui/service/mcp/McpSpecificationGenerator.java
+++ b/api/src/main/java/io/kafbat/ui/service/mcp/McpSpecificationGenerator.java
@@ -151,7 +151,7 @@ public class McpSpecificationGenerator {
   private Mono<CallToolResult> reponseToCallResult(ResponseEntity<?> response) {
     HttpStatusCode statusCode = response.getStatusCode();
     if (statusCode.is2xxSuccessful() || statusCode.is1xxInformational()) {
-      return Mono.just(this.callToolResult(response.getBody()));
+      return this.toCallResult(response.getBody());
     } else {
       try {
         return Mono.just(toErrorResult(objectMapper.writeValueAsString(response.getBody())));
--- a/api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
+++ b/api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
@@ -11,17 +11,21 @@ import com.github.victools.jsonschema.generator.SchemaGenerator;
 import com.github.victools.jsonschema.generator.SchemaGeneratorConfigBuilder;
 import com.github.victools.jsonschema.generator.SchemaVersion;
 import io.kafbat.ui.config.ClustersProperties;
+import io.kafbat.ui.controller.ClustersController;
 import io.kafbat.ui.controller.TopicsController;
 import io.kafbat.ui.mapper.ClusterMapper;
+import io.kafbat.ui.model.ClusterDTO;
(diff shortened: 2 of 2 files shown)


kanchan78:
Closing in favor of #1952 — opened a fresh PR from a branch named per the contributing guide's convention (`issues/1454`) with the full PR template filled in. Same fix, no functional changes.

PR #1952, opened 2026-08-23
Fix Flux serialization in ResponseEntity for MCP tools

### LINKED ISSUE #1454: MCP server  doesn't properly handle Flux responses.

kanchan78:

Fixes #1454.

`McpSpecificationGenerator.toCallResult()` already handles a bare `Flux<?>` correctly by calling `.collectList()` before serializing. But controllers like `ClustersController#getClusters`, `AclsController#listAcls`, `KafkaConnectController#getAllConnectors`, and `MessagesController#getTopicMessagesV2` return `Mono<ResponseEntity<Flux<T>>>` — the `Flux` is nested *inside* the `ResponseEntity` body.

`reponseToCallResult()` unwraps the `ResponseEntity` and passes its body straight to `callToolResult()`, skipping the `Flux`-aware branch in `toCallResult()` entirely. Jackson then introspects the live, unresolved `Flux` object graph and serializes its internal Reactor fields instead of its contents, e.g. `{"scanAvailable":true,"prefetch":-1}` instead of the actual list of clusters/ACLs/connectors/messages. The tool call reports `isError:false`, so an MCP client gets silently wrong data rather than a visible failure.

Fix: `reponseToCallResult()` now delegates the response body to `toCallResult()` instead of calling `callToolResult()` directly, so the existing `Flux` branch (`flux.collectList()`) applies uniformly whether the `Flux` is top-level or nested inside a `ResponseEntity`.

None — it's a one-line change (see diff), the risk is in the surrounding reasoning rather than the code itself, which the added regression test should make easy to check.

- [ ] No need to
- [x] Unit checks
- [ ] Integration checks

Added `toolBackedByResponseEntityOfFluxReturnsCollectedDataNotReactorInternals`, exercising `ClustersController#getClusters` through the real `McpSpecificationGenerator`. Verified it fails on unpatched code (reproduces `{"scanAvailable":...}`) and passes after the fix. Also manually reproduced the bug against a live deployment across `getClusters`, `listAcls`, `getAllConnectors`, and `getTopicMessagesV2` (all `ResponseEntity<Flux<T>>>`-shaped, all affected) before the fix, and confirmed `getTopics`/`getConsumerGroupsPage`/`getSchemas`/`getLatestSchema` (which return `Page<T>`/plain DTOs, not `Flux`) were unaffected — confirming the bug is specific to the `ResponseEntity<Flux<T>>` shape. After the fix, built the full application (including frontend) and smoke-tested the resulting image standalone: opened a real MCP SSE session and called `getClusters`, which returned `[]` (correct, no clusters configured) instead of the Reactor-internals payload.

🦫

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed MCP tool responses backed by reactive data streams to return collected data correctly instead of exposing internal serialization details.
  * Improved consistency when processing successful API responses.

### DIFF
--- a/api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
+++ b/api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
@@ -11,17 +11,21 @@ import com.github.victools.jsonschema.generator.SchemaGenerator;
 import com.github.victools.jsonschema.generator.SchemaGeneratorConfigBuilder;
 import com.github.victools.jsonschema.generator.SchemaVersion;
 import io.kafbat.ui.config.ClustersProperties;
+import io.kafbat.ui.controller.ClustersController;
 import io.kafbat.ui.controller.TopicsController;
 import io.kafbat.ui.mapper.ClusterMapper;
+import io.kafbat.ui.model.ClusterDTO;
 import io.kafbat.ui.model.KafkaCluster;
 import io.kafbat.ui.model.SortOrderDTO;
 import io.kafbat.ui.model.TopicColumnsToSortDTO;
 import io.kafbat.ui.model.TopicUpdateDTO;
+import io.kafbat.ui.service.ClusterService;
 import io.kafbat.ui.service.ClustersStorage;
 import io.kafbat.ui.service.KafkaConnectService;
 import io.kafbat.ui.service.TopicsService;
 import io.kafbat.ui.service.acl.AclsService;
 import io.kafbat.ui.service.analyze.TopicAnalysisService;
+import io.kafbat.ui.service.rbac.AccessControlService;
 import io.modelcontextprotocol.server.McpAsyncServerExchange;
 import io.modelcontextprotocol.server.McpServerFeatures.AsyncToolSpecification;
 import io.modelcontextprotocol.spec.McpSchema;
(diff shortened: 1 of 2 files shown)


api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
```diff
@@ -231,6 +235,37 @@ void writeOperationWithoutClusterNameReturnsError() {
         .verifyComplete();
   }
 
+  @Test
+  void toolBackedByResponseEntityOfFluxReturnsCollectedDataNotReactorInternals() {
```
Haarolean, line 239: a comment block narrating the bug and assertions on reactor internals. drop it, the whole test

Haarolean (changes requested):

Haarolean:
dupe of #1491

api/src/test/java/io/kafbat/ui/service/mcp/McpSpecificationGeneratorTest.java
kanchan78: Done — dropped the narrating comment block and the assertions coupled to Reactor's internal field names, kept the test focused on the actual contract (real data present, not an error). Pushed to the branch. Since this is closed as a dupe of #1491, mainly leaving it here for the record / in case it's useful.

PR #1978, opened 2026-09-04
Add "Not consumed by group" polling mode to message browser

turkcankeskin:

Adds a new polling mode to the topic message browser that shows the messages a consumer group has **not consumed yet**, so the backlog behind a lag can be inspected without manually looking up committed offsets and typing them into "From offset".

The mode is picked from the existing polling mode dropdown on **Topics → \<topic\> → Messages** as **"Not consumed by group"**. Selecting it reveals a consumer group picker (populated from the topic's own consumer groups); polling starts at that group's committed offset on every partition and runs forward to the end of the topic. All the other controls — partition filter, serdes, string/smart filters, export, cursor pagination — keep working unchanged.

Contract:
- New `PollingMode` value: `FROM_CONSUMER_GROUP_OFFSET`.
- New optional `consumerGroupId` query parameter on `GET /api/clusters/{clusterName}/topics/{topicName}/messages/v2`, required for that mode (a `ValidationException` is raised otherwise).

Backend (`MessagesService#resolveConsumerPosition`):
- Reads the group's committed offsets via `ReactiveAdminClient#listConsumerGroupOffsets` for the target partitions and rebuilds the `ConsumerPosition` with a per-partition offsets map.
- Partitions the group has no committed offset for are read from the very beginning — from that group's point of view nothing on them has been consumed.
- From there the existing `ForwardEmitter` + cursor pagination path is reused as-is; the new mode only needed to be added to the existing `SeekOperations` / `Cursor` switches, where it behaves exactly like `FROM_OFFSET` (a cursor for the next page is registered as `FROM_OFFSET`, so paging works the same way).
- RBAC: when `consumerGroupId` is passed, `ConsumerGroupAction.VIEW` on that group is required on top of `TopicAction.MESSAGES_READ`.

Frontend:
- New `ConsumerGroupSelect` component, rendered only for this mode and behind its own `Suspense` boundary so picking the mode doesn't blank out the whole Messages tab.
- The selected group is kept in the URL search params and in the per-topic filter local storage, like the other filters, and is cleared when switching to another mode.
- No request is fired until a group is picked, so the mode doesn't produce a 400 while the user is still choosing.

- `MessagesService#loadMessages`: `publishOn(Schedulers.boundedElastic())` had to move *after* the offset resolution. The resolution is an async admin-client call, so without the reorder the (blocking) emitter would be subscribed to on a Kafka admin client thread instead of the elastic scheduler.
- The choice to fall back to offset `0` for partitions with no committed offset. `SeekOperations#fixOffsets` clamps it up to the real beginning offset, so it is safe on compacted/retained topics, but it is a product decision worth confirming.
- Whether `FROM_CONSUMER_GROUP_OFFSET` is the right name, vs. something closer to the UI wording ("not consumed by group").

- [ ] No need to
- [x] Unit checks
- [x] Integration checks

Manually: ran the built image against a local Kafka, produced messages to a topic, committed a partial offset for a consumer group, and confirmed the mode lists exactly the messages after the committed offset, that the lag shrinks as the group catches up, and that the partition filter and pagination still apply.

Integration (`MessagesServiceTest`, 3 new cases):
- `loadMessagesFromConsumerGroupOffsetReturnsOnlyNotConsumedMessages` — 10 messages, offset 6 committed, expects `message_6`…`message_9`.
- `loadMessagesFromConsumerGroupOffsetReturnsAllMessagesWhenGroupHasNoCommittedOffsets`.
- `loadMessagesFromConsumerGroupOffsetReturnsExceptionWhenGroupNotProvided`.

Unit: extended `Filters/utils` specs for the new mode predicates; existing `SeekOperationsTest`, `CursorTest` and the frontend Messages/hooks suites pass unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a “Not consumed by group” message browsing mode.
  * Users can select a consumer group to view messages it has not yet consumed.
  * Consumer-group selection is preserved in the URL and filters.
  * Added API support for consumer-group offset polling.

* **Bug Fixes**
  * Improved handling of consumer-group offsets, including groups without committed offsets.
  * Added validation when a consumer group is required but not selected.

* **Documentation**
  * Documented the new polling mode in the feature list and API specifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ We extend our gratitude to Provectus for their past support in groundbreaking wo
 * **Metrics Dashboard** – Track key Kafka metrics in real time with a streamlined, lightweight dashboard.
 * **Kafka Brokers Overview** – Inspect brokers, including partition assignments and controller status.
 * **Consumer Group Details** – Analyze parked offsets per partition, and monitor both combined and partition-specific lag.
-* **Message Browser** – Explore messages in JSON, plain text, or Avro encoding formats. Live view is supported, enriched with user-defined CEL message filters.
+* **Message Browser** – Explore messages in JSON, plain text, or Avro encoding formats. Live view is supported, enriched with user-defined CEL message filters. Messages a consumer group has not consumed yet can be browsed with the "Not consumed by group" polling mode.
 * **Dynamic Topic Management** – Create and configure new topics with flexible, real-time settings.
 * **Pluggable Authentication** – Secure your UI using OAuth 2.0 (GitHub, GitLab, Google), LDAP, or basic authentication.
 * **Cloud IAM Support** – Integrate with **GCP IAM**, **Azure IAM**, and **AWS IAM** for cloud-native identity and access management.
--- a/api/src/main/java/io/kafbat/ui/controller/MessagesController.java
+++ b/api/src/main/java/io/kafbat/ui/controller/MessagesController.java
@@ -22,6 +22,7 @@ import io.kafbat.ui.model.TopicMessageEventDTO;
 import io.kafbat.ui.model.TopicSerdeSuggestionDTO;
 import io.kafbat.ui.model.rbac.AccessContext;
 import io.kafbat.ui.model.rbac.permission.AuditAction;
+import io.kafbat.ui.model.rbac.permission.ConsumerGroupAction;
 import io.kafbat.ui.model.rbac.permission.TopicAction;
 import io.kafbat.ui.serde.api.Serde;
 import io.kafbat.ui.service.DeserializationService;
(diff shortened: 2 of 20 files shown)


Haarolean:
Hi @turkcankeskin, thanks for the PR — the implementation itself is reasonably put together, so I want to be upfront that this isn't about code quality.

Per our [contributing guide](https://ui.docs.kafbat.io/development/contributing), features need to be backed by justified requirements and go through a triaged issue **before** implementation. There's no issue for this, nobody has requested this functionality in either this repo or the predecessor project, and it wasn't discussed with maintainers beforehand.

This matters more than usual here because the PR adds public API surface — a new `PollingMode` enum value and a `consumerGroupId` query parameter. Once that ships, it's a permanent contract we have to maintain, so the design (whether this should exist at all, whether it's an API mode vs. a UI entry point from the consumer group page, what it should be called) has to be agreed on first, not retrofitted in review. As an example: the "Not consumed by group" label is already a problem — committed offsets don't tell you what a group has actually *consumed*, only what it has committed — and that's exactly the kind of thing issue triage exists to catch.

I'm closing this for now. If you believe there's demand for browsing a consumer group's backlog, please open a feature request describing the use case, and we'll triage it. If it's accepted, this PR can be reopened and reviewed against the agreed design — the work wouldn't be wasted.

